PR #2, opened 2018-01-11
DEP: The Dat Enhancement Proposal Process (for Draft status)

bnewbold:
View it rendered [here](https://github.com/bnewbold/dat-deps/blob/meta-dep/proposals/0001-dep-process.md)

Also rendered [README](https://github.com/bnewbold/dat-deps/blob/meta-dep/README.md) and [template](https://github.com/bnewbold/dat-deps/tree/meta-dep).

This first DEP is now ready for review. I'm submitting it for Draft status, with the hope that it stabilizes to Active within a couple months after a couple other DEPs go through the process.

I believe I pulled in all the decisions from today's WG meeting ([notes](https://github.com/datprotocol/working-group/blob/master/meeting-notes/02-24jan2018.md)).

While we mentioned a two week review period during the meeting, I remembered that that's a *maximum*, so if everybody likes what they see here we could merge-as-draft sooner (though not less than ~3 days to give everybody time to read). Also totally fine to wait two weeks until the next meeting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,16 +8,17 @@
 This repository contains a series of "Dat Enhancement Proposal" (DEP)
 documents, part of the dat protocol development and standardization process.
 
-Accepted proposals can be viewed at [datprotocol.com](https://datprotocol.com).
+Accepted proposals will be viewable at [datprotocol.com](https://datprotocol.com).
 New (draft) proposals and discussion can be viewed on [github][github-deps]
 under "issues" and "pull-requests".
 
 These documents might be interesting reading to anybody wanting to learn more
 about protocol nitty gritties or the consensus process, but the documentation
-at [https://docs.datproject.org]() is specifically written with end-users and
+at [docs.datproject.org][docs] is specifically written with end-users and
 application developers in mind.
 
 [github-deps]: https://github.com/datprotocol/DEPs/
+[docs]: https://docs.datproject.com
 
 ## The Process
 
--- a/README.md
+++ b/README.md
@@ -1,3 +1,45 @@
+
 # Dat Enhancement Proposals
 
-Dat Enhancement Proposals. Contains all specs for the Dat protocol, including drafts.
+[![#dat IRC channel on freenode](https://img.shields.io/badge/irc%20channel-%23dat%20on%20freenode-blue.svg)](http://webchat.freenode.net/?channels=dat)
+[![datproject/discussions](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/datproject/discussions?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![docs](https://img.shields.io/badge/Dat%20Project-Docs-green.svg)](http://docs.dat-data.com)
+
+This repository contains a series of "Dat Enhancement Proposal" (DEP)
+documents, part of the dat protocol development and standardization process.
+
+Accepted proposals can be viewed at [datprotocol.com](https://datprotocol.com).
+New (draft) proposals and discussion can be viewed on [github][github-deps]
+under "issues" and "pull-requests".
+
+These documents might be interesting reading to anybody wanting to learn more
+about protocol nitty gritties or the consensus process, but the documentation
+at [https://docs.datproject.org]() is specifically written with end-users and
+application developers in mind.
+
+[github-deps]: https://github.com/datprotocol/DEPs/
+
+## The Process
+
+TL;DR: write up an DEP markdown file; submit a github PR; revise per review;
+community will either accept (merges PR) or not (closes PR). See the ["DEP
+Process"][dep-0001] meta-DEP for details.
+
+For now this process is as simple and informal; as the number of dat developers
+and users grows, it may become more formal. Precedents and inspiration include:
+
+- [Bittorrent Enhancement Proposals (BEP)](http://bittorrent.org/beps/bep_0001.html)
+- [Rust RFC Process](https://github.com/rust-lang/rfcs)
+- [Python Enhancement Proposal](https://www.python.org/dev/peps/pep-0001/)
+
+[dep-0001]: ./proposals/0001-dep-process.md
+
+## Fine Print
+
+The contents of this repository are all under a CC-BY (Creative Commons
+Attribution 4.0) license unless otherwise noted.
+
+Contributors and discussion participants are expected to abide by the Dat
+Project [Code of Conduct][coc].
+
+[coc]: https://github.com/datproject/Code-of-Conduct
--- a/README.md
+++ b/README.md
@@ -8,9 +8,7 @@
 This repository contains a series of "Dat Enhancement Proposal" (DEP)
 documents, part of the dat protocol development and standardization process.
 
-Accepted proposals will be viewable at [datprotocol.com](https://datprotocol.com).
-New (draft) proposals and discussion can be viewed on [github][github-deps]
-under "issues" and "pull-requests".
+Accepted proposals will be viewable at [datprotocol.com](https://datprotocol.com).  New (draft) proposals and discussion can be viewed on [github][github-deps] under "Pull requests".
 
 These documents might be interesting reading to anybody wanting to learn more
 about protocol nitty gritties or the consensus process, but the documentation
--- a/proposals/0001-dep-process.md
+++ b/proposals/0001-dep-process.md
@@ -9,10 +9,9 @@ Status: Draft (as of 2018-01-15)
 
 Github PR: [Draft](https://github.com/datprotocol/DEPs/pull/2)
 
-Authors: Dat Protocol Working Group: Danielle, Mattias, Joe Hand, Karissa,
-Paul, Tara, Bryan
-
-[Bryan Newbold](https://github.com/bnewbold)
+Authors: Dat Protocol Working Group: [Tara](https://github.com/taravancil),
+[Paul](https://github.com/pfrazee), [Mathias](https://github.com/mafintosh), [Karissa](https://github.com/karissa), [Joe](https://github.com/karissa), [Danielle](#),
+[Bryan](https://github.com/bnewbold)
 
 # Summary
 [summary]: #summary
--- a/proposals/0001-dep-process.md
+++ b/proposals/0001-dep-process.md
@@ -53,13 +53,14 @@ protocol. The overall developent and decision making process should be more
 transparent, accessible, and scalable to a growing group of application
 developers and end users
 
-# Submitting a Proposal
+
+# How To Submit a Proposal
 [submit]: #submit
 
 As a first step, before drafting a DEP or implementing experimental new
 protocol features, it's helpful to informally pitch your idea to see if others
 in the community are already thinking something similar, or have discussed the
-same idea in the past. This discussion could happen over chat, github issues,
+same idea in the past. This discussion could happen over chat, Github issues,
 blog posts, or other channels. If you can recruit collaborators and work out
 some of the details, all the better. This period could be called **pre-DEP**.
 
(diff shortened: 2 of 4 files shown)


proposals/0001-dep-process.md
```diff
@@ -0,0 +1,147 @@
...
+
+**TODO:**
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Who are "core developers"? What is the specific decision making process for
+accepting or rejecting a given DEP? Optimistically, it would be clear from
+reading a PR discussion thread whether "consensus" has been reached or not, but
+this might be ambiguous or intimidating to first-time contributors.
```
pfrazee: I propose we create a process whereby we establish "repo admins." These admins would be re-evaluated every N months (I propose 6), and being selected as an admin means committing to involvement for the full N months. At the end of the period, new "repo admins" are selected by unanimous consent of the previous period's admins, and that may mean no change in the admins. I would suggest one admin per organization which has an interest in Dat (which at this point would be Code for Science and Society, Blue Link Labs, and the Internet Archive) but there's no reason to follow that exactly.

For decision-making, I would also suggest that any decision requires unanimous consent from the repo admins, with an informal requirement that we factor community feedback in the decision.

proposals/0001-dep-process.md
```diff
@@ -0,0 +1,147 @@
...
+
+When are or are not accepted DEPs binding? Presumably a technical DEP should indicate
+
+Are "standard"/technical and "process" the right categories and terms to use?
+Some DEP systems (like the IETF one) also use terminology like "informational"
+(seems like we could rely on blog posts) an "normative". This might be helpful
+for documenting things like peer discovery methods, which have intentially not
+been specified as part of the formal spec (to make it clear that there is no
+single "right way" to do discovery), but they should be documented somewhere.
```
pfrazee: I think "normative" would be a good category to have, for the reasons you say.

proposals/0001-dep-process.md
```diff
@@ -0,0 +1,147 @@
...
+
+Some communities eventually subsume the core protocol specification itself into
+the DEP process (for example, the official Bittorrent protocol was documented
+in a "BEP"). Should dat do this? Or rely on the current (and possibly future)
+Dat whitepaper?
```
pfrazee: I vote that we put the protocol documentation into the DEP process.

pfrazee:
Nice start @bnewbold

proposals/0001-dep-process.md
bnewbold: Thinking outloud:
I'm not sure whether we should do this as one big document, or split it into components (basically along the node module abstraction layers: sleep, hypercore, hyperdb, discovery). I think the later... and then either have a "how it all fits together" DEP (maybe a "normative" one showing one way it could be done?), or rely on the whitepaper, or hand wave that it's self obvious.
I guess we don't actually need to show how a client would work all together in the specification, and should leave that as undefined as possible?

proposals/0001-dep-process.md
bnewbold: I didn't know how the word was defined when I wrote that. From Wikipedia:

> Normative = prescriptive = how to comply
> Informative = descriptive = help with conceptual understanding

I don't think that's the axis we're talking about here... more formal/informal? I think "Standard" should be more formal and strict ("if you're not doing it this way, You're Doing It Wrong"), and what we're looking for with a third category is "here's what some folks are doing and it works for them".

I fear "Normative" is overloaded. "Informative" probably works. Other similar words: convention, custom, descriptive

I think I like "Descriptive" best so far.

proposals/0001-dep-process.md
```diff
@@ -0,0 +1,147 @@
...
+  ammended based on review and comments. The PR will be tagged to indicate
+  DEP type and status.
+* Build consensus. This part of the process is not bounded in time, and will
+  likely involve both discussion on the PR comment thread and elsewhere (IRC,
+  etc).
+* Consider drafting or prototyping an implementation to demonstrate your
+  proposal and work out all the details. This step isn't necessary, however: a
+  proposer does not need to be a developer.
+* If consensus seems to have emerged (for or against the proposal), a team
+  member will assign an DEP number, update the status, and merge the PR.
```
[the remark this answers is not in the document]
bnewbold: Sure, I think that's reasonable. I think we can also have a lower bar for edits to a draft (eg, allow larger changes to the draft in response to discussion or implementation) than something that has been accepted.

proposals/0001-dep-process.md
bnewbold: This is a reasonable proposal. I think I want to research this governance aspect of the prospect a bit more... I will revise the other aspects, and maybe we can merge this as a draft to separate the desire to move ahead from committing to all the governance details?

proposals/0001-dep-process.md
pfrazee: Informative or Descriptive seem good to me

proposals/0001-dep-process.md
joehand: > I guess we don't actually need to show how a client would work all together in the specification, and should leave that as undefined as possible?

Seems like maybe this would fit in better at docs.datproject.org. It can be more hand waving so its easier to keep updated. Otherwise we may have to update with each spec change.

I think the key is that implementation details need to always be 100% up to date. More general, this is how Dat works don't need describe all the details.

pfrazee: I like the proposal of having DEPs for each abstraction layer and then an Informative/Descriptive on how they integrate. I think it does need to be a DEP though, because it's pretty important information for somebody making a full implementation. What I'd expect is that the "Fit-together DEP" would give an overview of the different pieces and describe how they interact, which would make it a very useful highlevel description of the protocol. Isolating the subDEPs would help them be specced more completely, and avoid enormous and unwieldly DEP docs.

proposals/0001-dep-process.md
tristanls: This article had a lot of very specific recommendations/proposals for "building cooperation into the system" that may be worth considering: https://medium.com/@nayafia/the-problem-with-voting-8cff39f771e8

bnewbold:
This round of edits was more divergent (more ideas, more text) than I was hoping. Will probably do a more convergent (polish, simplify) edit and then "submit" this weekend so it has a couple days to sit before our working group meeting next Wednesday.

README.md
```diff
@@ -1,3 +1,46 @@
...
+[![#dat IRC channel on freenode](https://img.shields.io/badge/irc%20channel-%23dat%20on%20freenode-blue.svg)](http://webchat.freenode.net/?channels=dat)
+[![datproject/discussions](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/datproject/discussions?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+[![docs](https://img.shields.io/badge/Dat%20Project-Docs-green.svg)](http://docs.dat-data.com)
+
+This repository contains a series of "Dat Enhancement Proposal" (DEP)
+documents, part of the dat protocol development and standardization process.
+
+Accepted proposals will be viewable at [datprotocol.com](https://datprotocol.com).
+New (draft) proposals and discussion can be viewed on [github][github-deps]
+under "issues" and "pull-requests".
```
joehand: All DEP proposal discussion should happen in PRs so we don't get discussion fractured.

proposals/0001-dep-process.md
```diff
@@ -0,0 +1,312 @@
...
+The Dat Enhancement Proposal ("DEP") process is how the Dat open source
+community comes to (distributed) consensus around technical protocol
+enhancements and organizational process.
+
+# Motivation
+[motivation]: #motivation
+
+The Dat protocol is still a living standard. A transparent process is needed
+for community members to understand what changes are in the pipeline and how
+new ideas might come to fruition.
```
joehand: Preparing to handle protocol version changes/breaks is also a primary motivation =). Should be mentioned upfront.

proposals/0001-dep-process.md
```diff
@@ -0,0 +1,312 @@
...
+   this stage.
+1. While in draft status, proposals can be experimented with. Small corrections
+   and clarifications can be submitted by PR expect to be merged quickly if
+   they are reasonable and don't change the broad behavior or semantics of the
+   proposal; larger changes should be re-submitted as Superceding proposals.
+1. When it seems approrpriate, a PR can be submitted to upgrade the status of a
+   Draft to Active. At this time a final review will take place, with the
+   outcome being that a prosal stays a Draft or is Active. It's also possible
+   for a Draft to be Closed (usuall by a specific PR to propose this). This
+   review period is shorter (**2 weeks maximum**), as the relevant reviewers
```
joehand: Should we also make sure this includes a discussion at the next working group?

proposals/0001-dep-process.md
```diff
@@ -0,0 +1,312 @@
...
+For a Draft to migrate to Active, there is an expectation that the proposal has
+been demonstrated, that the change of significant unforseen issues in complete
+adoption is low, and that the proposal will be the "new normal" and expected
+behavior going forward.
+
+
+# Decision Making Process
+[power]: #power
+
+There exists a Protocol Working Group (WG) which makes DEP status decisions.
```
joehand: Link to https://github.com/datprotocol/working-group

proposals/0001-dep-process.md
bnewbold: That sounds reasonable. I included a time-bound so PRs wouldn't get "stuck", but it could be a bit longer to ensure there is time for a WG meeting. I can imagine that, at least to start, a common workflow is going to be the WG itself discussing and proposing that a DEP be stabilized to "Active" anyways, which would make the scheduling easier.

I am unsure how much WG-internals should laid out in this DEP... but I don't want to just kick those decisions down the road either.

proposals/0001-dep-process.md
bnewbold: Almost forgot this!

bnewbold:
While writing this, I thought of a better way to describe the difference between the expectations about "Draft"/"Active" here (and in practice with Rust, and I think Python) vs. current practice with Bittorrent BEPs: whether the proposals lead implementation, or the implementation leads proposals and standards. W3C is maybe a good example of both: sometimes (HTML5?) the standards process guides implementation (browsers), but very often the opposite is true (standards process is playing catch up).

Obviously we've got a bunch of implementation running around already, but I think it could be really helpful (and realistic) for proposals and documentation to lead implementation.

proposals/0001-dep-process.md
```diff
@@ -5,25 +5,30 @@ Short Name: `0001-dep-process`
...
 Authors: [Dat Protocol Working Group][wg]:
-[Tara](https://github.com/taravancil), [Paul](https://github.com/pfrazee),
-[Mathias](https://github.com/mafintosh), [Karissa](https://github.com/karissa),
-[Joe](https://github.com/karissa), [Danielle](#),
-[Bryan](https://github.com/bnewbold)
+[Tara Vancil](https://github.com/taravancil),
+[Paul Frazee](https://github.com/pfrazee),
+[Mathias Buus](https://github.com/mafintosh),
+[Karissa McKelvey](https://github.com/karissa),
+[Joe Hand](https://github.com/karissa),
```
joehand: https://github.com/joehand

proposals/0001-dep-process.md
```diff
@@ -5,25 +5,30 @@ Short Name: `0001-dep-process`
...
+[Danielle Robinson](#),
```
joehand: https://github.com/daniellecrobinson

joehand:
Looks great, thanks for updating this!

pfrazee:
I'm happy with this DEP's status.

PR #3, opened 2018-02-05
Draft: HyperDB

bnewbold:
This DEP is submitted for review for Draft status.

[Rendered pre-merge](https://github.com/bnewbold/dat-deps/blob/dep-hyperdb/proposals/0000-hyperdb.md)

A diagram showing the trie structure might be nice to help with learning, but maybe the specification isn't the right place for that. The examples are verbose, but I think are important to clarify exactly how things work. I think exact binary protobuf output (hexdump style?) could also be included for some example messages, to make encoding and endianness completely unambiguous.

### DIFF
--- a/proposals/0000-hyperdb.md
+++ b/proposals/0000-hyperdb.md
@@ -7,9 +7,12 @@ Type: Standard
 
 Status: Undefined (as of 2018-03-XX)
 
-Github PR: (add HTTPS link here after PR is opened)
+Github PR: [https://github.com/datprotocol/DEPs/pull/3](https://github.com/datprotocol/DEPs/pull/3)
 
-Authors: [Stephen Whitmore](https://github.com/noffle), [Bryan Newbold](https://github.com/bnewbold)
+Authors:
+[Bryan Newbold](https://github.com/bnewbold),
+[Stephen Whitmore](https://github.com/noffle),
+[Mathias Buus](https://github.com/mafintosh)
 
 
 # Summary
@@ -567,12 +570,12 @@ concerns are out of scope for this DEP.
 # Changelog
 [changelog]: #changelog
 
-As of March 2018, @mafintosh is leading development of a hyperdb nodejs module
-on [github](https://github.com/mafintosh/hyperdb), which is the basis for this
-DEP.
+As of March 2018, Mathias Buus (@mafintosh) is leading development of a hyperdb
+nodejs module on [github](https://github.com/mafintosh/hyperdb), which is the
+basis for this DEP.
 
-- 2017-12-06: @noffle publishes `ARCHITECTURE.md` overview in the
-  [hyperdb github repo][arch_md]
-- 2018-02-XX: First complete draft submitted for review
+- 2017-12-06: Stephen Whitmore (@noffle) publishes `ARCHITECTURE.md` overview
+  in the [hyperdb github repo][arch_md]
+- 2018-03-04: First draft for review
 
 [arch_md]: https://github.com/mafintosh/hyperdb/blob/master/ARCHITECTURE.md
--- a/proposals/0000-hyperdb.md
+++ b/proposals/0000-hyperdb.md
@@ -20,7 +20,7 @@ Authors:
 
 Hyperdb is an abstraction layer providing a general purpose distributed
 key/value store over the hypercore protocol. It is an iteration on the
-hyperdrive directory tree implementation, building top of the hypercore
+hyperdrive directory tree implementation, building on top of the hypercore
 append-only log abstraction layer. Keys are path-like strings (eg,
 `/food/fruit/kiwi`), and values are arbitrary binary blobs (generally under a
 megabyte).
@@ -99,9 +99,9 @@ or the "latest" revision.
 Requires read-write access.  Returns an error (eg, via callback) if there was a
 problem.
 
-`db.get(key)`: Reading a non-existant `key` is an error. Read-only.
+`db.get(key)`: Reading a non-existent `key` is an error. Read-only.
 
-`db.delete(key)`: Removes the key from the database. Deleting a non-existant
+`db.delete(key)`: Removes the key from the database. Deleting a non-existent
 key is an error. Requires read-write access.
 
 `db.list(prefix)`: returns a flat (not nested) list of all keys currently in
@@ -163,14 +163,15 @@ own DEP and mentioned only partially here. The fields common to both message
 types are:
 
 - `key`: UTF-8 key that this node describes. Leading and trailing forward
-  slashes (`/`) are always striped before storing in protobuf.
-- `value`: arbitrary byte array. A non-existant `value` entry indicates that
+  slashes (`/`) are always stripped before storing in protobuf.
+- `value`: arbitrary byte array. A non-existent `value` entry indicates that
   this Entry indicates a deletion of the key; this is distinct from specifying
   an empty (zero-length) value.
 - `trie`: a structured array of pointers to other Entry entries in the feed,
   used for navigating the tree of keys.
-- `clock`: reserved for use in the forthcoming `multi-writer` standard An empty
-  list is the safe (and expected) value for `clock` in single-writer use cases.
+- `clock`: reserved for use in the forthcoming `multi-writer` standard. An
+  empty list is the safe (and expected) value for `clock` in single-writer use
+  cases.
 - `inflate`: a "pointer" (reference to a feed index number) to the most recent
   entry in the feed that has the optional `feeds` and `contentFeed` fields set.
   Not defined if `feeds` and `contentFeed` are defined.
@@ -218,7 +219,7 @@ case we use all zeros.
 When converting the 8-bytehash to an array of 2-bit bytes, the ordering is
 proceed byte-by-byte, and for each take the two lowest-value bits (aka, `hash &
 0x3`) as byte index `0`, the next two bits (aka, `hash & 0xC`) as byte index
-`1`, etc. When concatanating path hashes into a longer array, the first
+`1`, etc. When concatenating path hashes into a longer array, the first
 ("left-most") path element hash will correspond to byte indexes 0 through 31;
 the terminator (`4`) will have the highest byte index.
 
@@ -323,7 +324,7 @@ always 0, so we consider only the entry index.
 For a `trie` with `N` buckets, each may have zero or more pointers. Typically
 there are a maximum of 3 pointers per bucket, corresponding to the 4 possible
 values minus the current Entry's value, but in the event of hash collisions (in
-the path array space) , there may be multiple pointers in the same bucket
+the path array space), there may be multiple pointers in the same bucket
 corresponding to the same value.
 
 To lookup a key in the database, the recipe is to:
@@ -359,7 +360,7 @@ Similarly, to write a key to the database:
    a pointer to the Entry with the same hash at the final array index.
 5. If the paths don't entirely match, find the first index at which the two
    arrays differ. Copy all `trie` elements (empty or not) into the new `trie`
-   for indicies between the "current index" and the "differing index".
+   for indices between the "current index" and the "differing index".
 6. Next look up the corresponding element in this Entry's `trie` array at the
    differing index. If this element is empty, then you have found the most
    similar Entry. Write a pointer to this node to the `trie` at
@@ -382,7 +383,7 @@ arrays of pointers to entries) into a variable-length bytestring as the `trie`
 field of an Entry protobuf message.
 
 Consider a trie array with `N` buckets and `M` non-empty buckets (`0 <= M <=
-N`). In the encoded field, there will be `M` concatanated bytestrings of the
+N`). In the encoded field, there will be `M` concatenated bytestrings of the
 form:
 
 - trie index (varint), followed by...
@@ -449,9 +450,9 @@ Now `M` is 2. The first bytestring chunk will be:
 the second bytestring chunk would be:
 
 - index varint is `64` (65th element in the trie array; the terminating value)
-- btifield is `0b10000` (varint 1); there is a single pointer set... but it
+- bitfield is `0b10000` (varint 1); there is a single pointer set... but it
   contains a hash collision, so there are two pointers
-- first pointer is (`feed=0` << 1 | 1`, `index=23`) or (varint 1, varint=23);
+- first pointer is (`feed=0 << 1 | 1`, `index=23`) or (varint 1, varint=23);
   note the `more` bit is set high!
 - second pointer is (`feed=1 << 1 | 0`, `index=17`) or (varint 2, varint 17);
   note the `more` bit is low, as usual. In the extremely unlikely case of
@@ -547,7 +548,7 @@ pointer to entry index 1, so we fetch that Entry and recurse. Comparing path
 hash arrays, we now get all the way to index 34 before there is a difference.
 We again look in the `trie`, find a pointer to entry index 0, and fetch the
 first Entry and recurse. Now the path elements match exactly; we have found the
-Entry we are looking for, and it has an existant `value`, so we return the
+Entry we are looking for, and it has an existent `value`, so we return the
 `value`.
 
 Consider a lookup for `db.get('/a/z')`; this key does not exist, so we expect
@@ -658,7 +659,7 @@ The total metadata overhead for a database with M entries scales with `O(M
 
 The basic key/value semantics of hyperdb (as discussed in this DEP, not
 considering multi-writer changes) are not known to introduce new privacy issues
-when compared with, eg, storing binary values at key-like paths using the
+when compared with, e.g., storing binary values at key-like paths using the
 current ("legacy") hyperdrive system.
 
 A malicious writer could cause trouble for readers, even readers who do not
@@ -699,7 +700,7 @@ for these fields.
 
 Hyperdb is not backwards compatible with the existing hyperdrive metadata,
 meaning dat clients may need to support both versions during a transition
-period. This applies both to archives saved to disk (eg, in SLEEP) and to
+period. This applies both to archives saved to disk (e,g., in SLEEP) and to
 archives received and published to peers over the network.
 
 No changes to the Dat network wire protocol itself are necessary, only changes
@@ -721,7 +722,7 @@ Should we declare that `contendFeed` pointers *not* change over the history of
 a feed? See also <https://github.com/datprotocol/DEPs/issues/13>
 
 Need to think through deletion process with respect to listing a path prefix;
-will previously deleted nodes be occulded, or potentially show up in list
+will previously deleted nodes be occluded, or potentially show up in list
 results? Should be reviewed (by a non-author of this document) before accepted
 as a Draft.
 
--- a/proposals/0000-hyperdb.md
+++ b/proposals/0000-hyperdb.md
@@ -21,12 +21,12 @@ Authors:
 Hyperdb is an abstraction layer providing a general purpose distributed
 key/value store over the hypercore protocol. It is an iteration on the
 hyperdrive directory tree implementation, building on top of the hypercore
-append-only log abstraction layer. Keys are path-like strings (eg,
+append-only log abstraction layer. Keys are path-like strings (e.g.,
 `/food/fruit/kiwi`), and values are arbitrary binary blobs (generally under a
 megabyte).
 
 Hyperdrive (used by the Dat application) is expected to be re-implemented on
-top of hyperdb for improved performance with many files (eg, millions). The
+top of hyperdb for improved performance with many files (e.g., millions). The
 hyperdrive API should be largely unchanged, but the `metadata` format will be
 backwards-incompatible.
 
@@ -71,13 +71,13 @@ A Hyperdb database instance can be represented by a single hypercore feed (or
 several feeds in a multi-writer context), and is named, referenced, and
 discovered using the public and discovery keys of the hypercore feed (or the
 original feed if there are several). In a single-writer configuration, only a
-single node (holding the secret key) can mutate the database (eg, via `put` or
+single node (holding the secret key) can mutate the database (e.g., via `put` or
 `delete` actions).
 
 **Keys** can be any UTF-8 string. Path segments are separated by the forward
 slash character (`/`). Repeated slashes (`//`) are disallowed. Leading and
 trailing `/` are optional in application code: `/hello` and `hello` are
-equivalent. A key can be both a "path segment" and key at the same time; eg,
+equivalent. A key can be both a "path segment" and key at the same time; e.g.,
 `/a/b/c` and `/a/b` can both be keys at the same time.
 
 **Values** can be any binary blob, including empty (of zero length). For
@@ -96,7 +96,7 @@ creating a new one. A handle could represent any specific revision in history,
 or the "latest" revision.
 
 `db.put(key, value)`: inserts `value` (arbitrary bytes) under the path `key`.
-Requires read-write access.  Returns an error (eg, via callback) if there was a
+Requires read-write access.  Returns an error (e.g., via callback) if there was a
 problem.
 
 `db.get(key)`: Reading a non-existent `key` is an error. Read-only.
@@ -187,9 +187,9 @@ feed, with `feeds` containing a single entry (a pointer to the current feed
 itself), and `contendFeed` optionally set to a pointer to a paired contend
 feed.
 
-If *either* `feeds` *or* `contentFeed` are definined in an entry, *both* fields
-must be defined, so the new message can be refered to via `inflated`. In this
-case the entry is refered to as an "inflated entry".
+If *either* `feeds` *or* `contentFeed` are defined in an entry, *both* fields
+must be defined, so the new message can be referred to via `inflated`. In this
+case the entry is refereed to as an "inflated entry".
 
 
 ## Path Hashing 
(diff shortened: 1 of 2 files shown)


proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+Authors:
+[Bryan Newbold](https://github.com/bnewbold),
+[Stephen Whitmore](https://github.com/noffle),
+[Mathias Buus](https://github.com/mafintosh)
+
+
+# Summary
+[summary]: #summary
+
+HyperDB is a new abstraction layer providing a general purpose distributed
```
pfrazee: Is it "new"? Aint this all new?

pfrazee: (Not just nitpicking on prose; I'm concerned "new" will be confusing, both now and later when nothing is comparatively newer.)

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+key/value store over the Dat protocol. It is an iteration on the hyperdrive
+directory tree implementation, building top of the hypercore append-only log
+abstraction layer. Keys are path-like strings (eg, `/food/fruit/kiwi`), and
+values are arbitrary binary blobs (generally under a megabyte).
+
+Hyperdrive (used by the Dat application) is expected to be re-implemented on
+top of HyperDB for improved performance with many files (eg, millions). The
+hyperdrive API should be largely unchanged, but the `metadata` format will be
+backwards-incompatible.
```
pfrazee: Yeah I suggest we move this paragraph about hyperdrive into the motivation, where the plans and reasoning at the time of writing can all be contained.

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+HyperDB is expected to drastically improve performance of dat clients when
+working with archives containing tens of thousands of files in single
+directories. This is a real-world bottleneck for several current users, with
+basic local actions such as adding a directory taking an unacceptably long time
+to complete.
+
+A secondary benefit is to refactor the [trie][trie]-structured key/value API
+out of hyperdrive, allowing third party code to build applications directly on
+this abstraction layer.
+
```
pfrazee: Third motivation: add multi-writer support

pfrazee: Oh scratch that, multi-writer isn't discussed in this doc

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+or the "latest" revision.
+
+`db.put(key, value)`: inserts `value` (arbitrary bytes) under the path `key`.
+Requires read-write access.  Returns an error (eg, via callback) if there was a
+problem.
+
+`db.get(key)`: Reading a non-existant `key` is an error. Read-only.
+
+`db.delete(key)`: Removes the key from the database. Deleting a non-existant
+key is an error. Requires read-write access.
```
pfrazee: Does deletion of a path-segment cause the deletion of all keys which the segment is a prefix of to be deleted?

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+  an empty (zero-length) value.
+- `clock`: reserved for use in the forthcoming `multi-writer` standard, not
+  described here. An empty list is the safe (and expected) value for `clock` in
+  single-writer use cases.
+- `trie`: a structured array of pointers to other Node entries in the feed,
+  used for navigating the tree of keys.
+- `path`: a 2-bit hash sequence of `key`'s components.
+- `seq`: the zero-indexed sequence number of this Node (hyperdb) entry in
+  the feed. Note that this may be offset from the hypercore entry index if
+  there are any leading non-hyperdb entries.
```
pfrazee: Does this mean that `seq` only increments when a hyperdb Node is written to the feed?

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+for these fields.
+
+
+# Dat migration logistics
+[migration]: #migration
+
+HyperDB is not backwards compatible with the existing hyperdrive metadata,
+meaning dat clients may need to support both versions during a transition
+period. This applies both to archives saved to disk (eg, in SLEEP) and to
+archives received and published to peers over the network.
```
pfrazee: How can a client detect which metadata protocol to use?

bnewbold: Good question. This is related to the issue of hashbase (and other tools) choking when they try to download a non-hyperdrive hypercore.
Two long-term solutions I could imagine are:
- add a Feed-level "type" field to hypercore itself. The obvious place for this would be the Feed message, except this would then leak type information in clear text on the wire, which is bad. A new message type? I could imagine refactoring hypercore messages to have a new "connection init" message for crypto setup, then re-send a Feed encrypted. This would be a good place to toss in a protocol version header as well. hypercore has been pretty stable though, this would be disruptive. Also, (thinking out loud) how would the software figure out the type of feeds on disk (aka, mystery meat SLEEP files)?
-  extend the convention of always writing a "special" protobuf message as the first entry in every hypercore feed, and encode type info there. hyperdrive currently does this with a pointer to the data feed, but no other hypercore feed "types" do this (AFAIK).

@mafintosh, thoughts?

proposals/0000-hyperdb.md
bnewbold: Good question! I think "No", a separate `delete_recursive()` (or a flag) would be needed.
I've tried to minimize the API surface here in the spec; eg, I haven't included the diff or history streaming stuff. I think in this case it's pretty straightforward to combined `list(prefix)` and `delete(key)`.

proposals/0000-hyperdb.md
bnewbold: Good point; i'm still finding the right "voice" to use: tense, "we/I", "should/could/must", "one/you".

I was thinking of adding a CI pass with something like http://proselint.com/; we should probably have at least a pass that validates the markdown, and maybe checks out-links for valid HTTP responses.

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+implementation (or configuration) specific. Read-only.
+
+If the hypercore underlying a hyperdb is only partially replicated, behavior is
+implementation-specific. For example, a `get()` call could block until the
+relevant value is replicated, or the implementation could return an error.
+
+An example pseudo-code session working with a database might be:
+
+    db.put('/life/animal/mammal/kitten', '{"cuteness": 500.3}')
+    db.put('/life/plant/bush/bananna', '{"delicious": 103.4}')
```
dcposch: don't know if you care but the `banana` has an extra n

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+The data structure of the trie is a sparse array of pointers to other Node
+entries. Each pointer indicates a feed index and a sequence number; for the
+non-multi-writer case, the feed index is always 0, so we consider only the
+sequence number (aka, entry index).
+
+To lookup a key in the database, the recipe is to:
+
+1. Calculate the `path` array for the key you are looking for.
+2. Select the most-recent ("latest") Node for the feed.
+3. Compare path arrays. If the paths match exactly, you have found the Node you
```
dcposch: SipHash is not collision resistant, so two distinct paths could have the same path array, right?

mafintosh: Yes, in case of a collision the final node bucket will just have more than one node in it. The full final node bucket is read on every read (usually only a single node) to do key full key comparison to check for collisions

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+A major motivator for HyperDB is to improve scaling performance with tens of
+thousands through millions of files per directory in the existing hyperdrive
+implementation. The current implementation requires the most recent node in a
+directory to point to all other nodes in the directory. Even with pointer
+compression, this requires on the order of `O(N^2)` bytes; the HyperDB
+implementation scales with `O(N log(N))`.
```
dcposch: 👍 👍 

this is what prevented us from doing [datpedia](https://github.com/dcposch/datpedia) the straightforward way

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+simpler to use all-zeros, and potentially makes it easier to copy or migrate
+HyperDB content between hypercore feeds. Referencing the feed public key breaks
+abstraction layers and the separation of concerns.
+
+There are implied "reasonable" limits on the size (in bytes) of both keys and
+values, but they are not formally specified. Protobuf messages have a hard
+specified limit of 2 GByte (due to 32-bit signed arthimetic), and most
+implementations impose a (configurable) 64 MByte limit. Should this DEP impose
+specific limits on key and value sizes? Would be good to decide before Draft
+status.
```
dcposch: since dat does rabin chunking, the 2GB limit seems fine ~ large files will always be divided into chunks smaller than that.

will the chunks always be below 64MB though? @mafintosh

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+implementations behave if they detect a collision (matching path but not
+matching key)?
+
+Need to think through deletion process with respect to listing a path prefix;
+will previously deleted nodes be occulded, or potentially show up in list
+results?
+
+Are the "deletion" semantics here correct, in that deletion Nodes persist as
+"tombstones", or should deleted keys be omitted from future `trie` fields to
+remove their presence?
```
dcposch: it seems important to leave deletion nodes in place--that way, `list()` can iterate over all current and deleted nodes under a directory and then omit the deleted ones

otherwise, i think `list()` might return partial results

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,581 @@
...
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+How are hash collisions handled? Eg, two keys which have the same `path` hash
+array. How unlikely is this situation? Eg, how many keys in a database before
+the birthday paradox results in a realistic chance of a collision? How should
+implementations behave if they detect a collision (matching path but not
+matching key)?
```
dcposch: with 64-bit keys, directories with over 2^32 = ~4 billion files are likely to contain collisions

maybe let the trie list be one element longer than `path`. if the path matches but `key` does not, then use that last element in `trie` to point to the previous match

dcposch: or that might not even be necessary. in the very rare cases where `get` finds a node where the `path` matches what it's looking for but `key` does not match, it could just go to the parent of that node, then restart the search. in other words: iterate through `path` matches until there are none left or we find the `key` match

mafintosh: https://github.com/mafintosh/hyperdb/commit/fd6034421eb4d1763260d80a636b0e7cb17cba0a <-- i've added @dcposch's hash collision to the tests. We support hash collisions, the two nodes just end up in the same bucket similar to have a multiwriter fork would

proposals/0000-hyperdb.md
mafintosh: @dcposch thats how it works atm actually

proposals/0000-hyperdb.md
mafintosh: @dcposch note this actually not for the files but just for the file *metadata*. in our case that is just a fs.Stat object with an additional numeric pointer to another feed where the actual file content is stored. This gives basically as large files as you'd want

bnewbold: Also, FWIW, dat (via hyperdrive) does not actually use Rabin chunking at the moment, just fixed size chunks. The protocol is agnostic to chunking, so this can evolve in the future with little disruption if/when content-aware chunking is more performent.

bnewbold: To remind, dat (via hyperdrive) is only one possible application on top of hyperdrive; while hyperdrive should be pretty comfortable even with a couple KByte metadata size limit, other applications might want to store arbitrary data in the hyperdb values; setting expectations around this could save frustration or growing pains later.

proposals/0000-hyperdb.md
bnewbold: Thanks for the catch!

proposals/0000-hyperdb.md
mafintosh: @bnewbold agreed. imo you should always only store "small" values directly in the hyperdb, say < 1mb. anything you should put in a "content feed" and then link it from the value.

bnewbold:
This is now really close. I'd like @mafintosh to look over my trie encoding stuff really quick (or just review my comment https://github.com/mafintosh/hyperdb/issues/79#issuecomment-379616618) first. There is also some ambiguity over whether the `contentFeed` field can actually be dynamically updated, which is being discussed in https://github.com/datprotocol/DEPs/issues/13.

There are one or two small details (eg, it would be great to check my "big-O" estimate for get() performance), and i'm a little nervous about all the nitty gritties because I haven't actually *implemented* any of this in working code, but I don't think those are hard blockers for Draft status.

proposals/0000-hyperdb.md
bnewbold: I think for Draft we can mention the inherent limits and then mention:

"It is STRONGLY recommended to keep both keys and values to less than 2 megabytes (individually and combined). If larger value sizes are desired, a separate content can be pointed to (via `contentFeed`), and the value in this feed can be metadata pointing to data (eg, contiguous blocks of entries) in that feed."

proposals/0000-hyperdb.md
mafintosh: Supporting recursive deletes thru a single append is actually not hard and could be added in the future

bnewbold:
@dcposch I want to re-iterate how great it is that you discovered an actual hash collision! I will update the DEP to include them as examples so promote their use as tests in independent implementations.

bnewbold:
I'm pushing this for review now.

The issue of encoding "type" of a feed is still open, but I think we can handle that in a follow-on DEP.

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+[Stephen Whitmore](https://github.com/noffle),
+[Mathias Buus](https://github.com/mafintosh)
+
+
+# Summary
+[summary]: #summary
+
+Hyperdb is an abstraction layer providing a general purpose distributed
+key/value store over the hypercore protocol. It is an iteration on the
+hyperdrive directory tree implementation, building top of the hypercore
```
ralphtheninja: building `on top` of?

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+creating a new one. A handle could represent any specific revision in history,
+or the "latest" revision.
+
+`db.put(key, value)`: inserts `value` (arbitrary bytes) under the path `key`.
+Requires read-write access.  Returns an error (eg, via callback) if there was a
+problem.
+
+`db.get(key)`: Reading a non-existant `key` is an error. Read-only.
+
+`db.delete(key)`: Removes the key from the database. Deleting a non-existant
```
ralphtheninja: `non-existent`

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+      repeated Feed feeds = 6;
+      optional bytes contentFeed = 7;
+    }
+
+Some fields are are specific to the multi-writer features described in their
+own DEP and mentioned only partially here. The fields common to both message
+types are:
+
+- `key`: UTF-8 key that this node describes. Leading and trailing forward
+  slashes (`/`) are always striped before storing in protobuf.
```
ralphtheninja: `stripped`, `striped` means having stripes

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+- `value`: arbitrary byte array. A non-existant `value` entry indicates that
```
ralphtheninja: `non-existent`

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+  this Entry indicates a deletion of the key; this is distinct from specifying
+  an empty (zero-length) value.
+- `trie`: a structured array of pointers to other Entry entries in the feed,
+  used for navigating the tree of keys.
+- `clock`: reserved for use in the forthcoming `multi-writer` standard An empty
```
ralphtheninja: `standard. An empty`

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+16-byte key; the input is the UTF-8 encoded path string segment, without any
+`/` separators or terminating null bytes. An implementation of this hash
+algorithm is included in the libsodium library in the form of the
+`crypto_shorthash()` function. A 16-byte "secret" key is required; for this use
+case we use all zeros.
+
+When converting the 8-bytehash to an array of 2-bit bytes, the ordering is
+proceed byte-by-byte, and for each take the two lowest-value bits (aka, `hash &
+0x3`) as byte index `0`, the next two bits (aka, `hash & 0xC`) as byte index
+`1`, etc. When concatanating path hashes into a longer array, the first
```
ralphtheninja: concatenating

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+
+The data structure of the trie is a sparse array of pointers to other Entry
+entries. Each pointer indicates a feed index and an entry index pointer, and is
+associated with a 2-bit value; for the non-multi-writer case, the feed index is
+always 0, so we consider only the entry index.
+
+For a `trie` with `N` buckets, each may have zero or more pointers. Typically
+there are a maximum of 3 pointers per bucket, corresponding to the 4 possible
+values minus the current Entry's value, but in the event of hash collisions (in
+the path array space) , there may be multiple pointers in the same bucket
```
ralphtheninja: ),

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+2. Select the most-recent ("latest") Entry for the feed.
+3. Compare path hash arrays. If the paths match exactly, and the keys match, then
+   you are overwriting the current Entry, and can copy the "remainder" of it's
+   `trie` up to your current `trie` index.
+4. If the paths match, but not the keys, you are adding a new key to an
+   existing hash bucket. Copy the `trie` and extend it to the full length. Add
+   a pointer to the Entry with the same hash at the final array index.
+5. If the paths don't entirely match, find the first index at which the two
+   arrays differ. Copy all `trie` elements (empty or not) into the new `trie`
+   for indicies between the "current index" and the "differing index".
```
ralphtheninja: indices

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+
+## Binary Trie Encoding
+[trie_encoding]: #trie_encoding
+
+The following scheme is used to encode trie data structures (sparse, indexed
+arrays of pointers to entries) into a variable-length bytestring as the `trie`
+field of an Entry protobuf message.
+
+Consider a trie array with `N` buckets and `M` non-empty buckets (`0 <= M <=
+N`). In the encoded field, there will be `M` concatanated bytestrings of the
```
ralphtheninja: concatenated

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+- index varint is `1` (second element in the trie array)
+- bitfield is `0b01110` (varint 9): there are three pointer sets
+- first pointer set is (`feed=0 << 1 | 0`, `index=1`) or (varint 2, varint 1)
+- second pointer set is (`feed=5 << 1 | 0`, `index=3`) or (varint 10, varint 3)
+- third pointer set is (`feed=6 << 1 | 0`, `index=98`) or (varint 12, varint 98)
+
+the second bytestring chunk would be:
+
+- index varint is `64` (65th element in the trie array; the terminating value)
+- btifield is `0b10000` (varint 1); there is a single pointer set... but it
```
ralphtheninja: bitfield

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+successfully return `'24'`, as written in the first Entry). First we calculate
+the path for the key `a/b`, which will be the same as the first Entry. Then we
+take the "latest" Entry, with entry index 2. We compare the path hash arrays,
+starting at the first element, and find the first difference at index 1 (`1 ==
+1`, then `1 != 2`). We look at index 1 in the current Entry's `trie` and find a
+pointer to entry index 1, so we fetch that Entry and recurse. Comparing path
+hash arrays, we now get all the way to index 34 before there is a difference.
+We again look in the `trie`, find a pointer to entry index 0, and fetch the
+first Entry and recurse. Now the path elements match exactly; we have found the
+Entry we are looking for, and it has an existant `value`, so we return the
```
ralphtheninja: existent

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+The total metadata overhead for a database with M entries scales with `O(M
+* log(M))`.
+
+
+# Privacy and Security Considerations
+[privacy]: #privacy
+
+The basic key/value semantics of hyperdb (as discussed in this DEP, not
+considering multi-writer changes) are not known to introduce new privacy issues
+when compared with, eg, storing binary values at key-like paths using the
```
ralphtheninja: e.g.

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+in this specification to indicate the safe "single-writer-only" values to use
+for these fields.
+
+
+# Dat migration logistics
+[migration]: #migration
+
+Hyperdb is not backwards compatible with the existing hyperdrive metadata,
+meaning dat clients may need to support both versions during a transition
+period. This applies both to archives saved to disk (eg, in SLEEP) and to
```
ralphtheninja: e.g.

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Should we declare that `contendFeed` pointers *not* change over the history of
+a feed? See also <https://github.com/datprotocol/DEPs/issues/13>
+
+Need to think through deletion process with respect to listing a path prefix;
+will previously deleted nodes be occulded, or potentially show up in list
```
ralphtheninja: occluded?

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,762 @@
...
+An example pseudo-code session working with a database might be:
+
+    db.put('/life/animal/mammal/kitten', '{"cuteness": 500.3}')
+    db.put('/life/plant/bush/banana', '{"delicious": 103.4}')
+    db.delete('/life/plant/bush/banana')
+    db.put('/life/plant/tree/banana', '{"delicious": 103.4}')
+    db.get('/life/animal/mammal/kitten')
+    => {"cuteness": 500.3}
+    db.list('/life/')
+    => ['/life/animal/mammal/kitten', '/life/plant/tree/banana']
```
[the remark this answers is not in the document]
bnewbold: I intended for it to be recursive; looks like there is a `recursive` flag in maf's hyperdb library. I've added a clarification.

bnewbold:
@noffle I've still got you in here as a co-author in attribution of your ARCHITECTURE.md document that was very helpful when starting this DEP. However, while this is derivative, almost none of that document remains in here, and I don't want to "co-author" you without consent. Any thoughts? I'll move you down to an Acknowledgements section if I don't hear back in a day or two.

noffle:
@bnewbold whatever you think is appropriate is fine by me!

bnewbold:
Any further comments here? Poke @mafintosh. If we could approve/merge this week that would be great (eg, at the protocol WG meeting next Wednesday at the latest).

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,764 @@
...
+feed from scratch, meaning new public/private key pairs, and that public key
+URL links will need to change.
+
+Further logistical details are left to the forthcoming Multi-Writer DEP.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+Should we declare that `contendFeed` pointers *not* change over the history of
```
mafintosh: yes

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,764 @@
...
+will previously deleted nodes be occluded, or potentially show up in list
+results? Should be reviewed (by a non-author of this document) before accepted
+as a Draft.
+
+Can the deletion process (currently leaving "tombstone" entries in the `trie`
+forever) be improved, such that these entries don't need to be iterated over?
+mafintosh mentioned this might be in the works. Does this DEP need to "leave
+room" for those changes, or should we call out the potential for future change?
+Probably not, should only describe existing solutions. This can be resolved
+after Draft.
```
mafintosh: it can and should be improved in the future (backwards compat). step one is shipping what we have

proposals/0000-hyperdb.md
```diff
@@ -0,0 +1,764 @@
...
+
+Review (or prove?) the `O(log(M))` intuition for `get()` operations. This could
+happen after Draft status.
```
mafintosh: it's log(M) yes

bnewbold:
To be explicit: left @noffle as a co-author. Thanks for kicking this whole thing off!

PR #4, opened 2018-02-05
Hypercore DEP - Ready for draft

pfrazee:
Starting this tracking PR now.

### DIFF
--- a/proposals/0000-hypercore.md
+++ b/proposals/0000-hypercore.md
@@ -23,9 +23,9 @@ Dat uses two registers, `content` and `metadata`. The `content` register contain
 # Motivation
 [motivation]: #motivation
 
-Many datasets are shared online today using HTTP and FTP, which lack built in support for version control or content addressing of data. This results in link rot and content drift as files are moved, updated or deleted. Dat solves this with a distributed, versioned file-sharing network which enables multiple devices to act in concert as a single virtual host.
+Many datasets are shared online today using HTTP, which lacks built in support for version control or content addressing of data. This results in link rot and content drift as files are moved, updated or deleted. Dat solves this with a distributed, versioned file-sharing network that enables multiple untrusted devices to act as a single virtual host.
 
-To ensure files are hosted correctly by untrusted devices, Dat needs a data structure which verifies the integrity of content and which retains a history of revisions. Notably, the data structure must:
+To ensure files are hosted correctly and can be referenced at different points in time, Dat needs a data structure which verifies the integrity of content and which retains a history of revisions. Notably, the data structure must:
 
  - Provide verification of file integrity using only the dataset identifier
  - Retain a full history of revisions to the dataset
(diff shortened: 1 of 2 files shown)


pfrazee:
Current draft has content extracted from the whitepaper and a little bit of commentary from me (drawbacks, alternatives, unresolved questions).

proposals/0000-hypercore.md
```diff
@@ -0,0 +1,274 @@
...
+
+8 = h(chunk4)
+9 = h(8 + 10)
+10 = h(chunk5)
+```
+
+In the Hypercore feed, we only want one active root. Therefore, when there are multiple roots we hash all the roots together again. We also include a big endian uint64 binary representation of the corresponding node index (TODO: why?). At most there will be `log2(number of data blocks)`.
+
+```
+root = h(uint64be(#9) + 9 + uint64be(#3) + 3)
```
mafintosh: We encode the index also as that describes the structure of the tree not just the data. This might not be nessesary but my gut feeling is that you could do a tree preimage attack on it https://en.m.wikipedia.org/wiki/Merkle_tree#Second_preimage_attack. We already defend against the preimage using the hash 1 byte prefixes so this might not be nessesary here but it doesn't hurt

mafintosh: It basically very strongly encodes the length of the feed in the hash

proposals/0000-hypercore.md
```diff
@@ -0,0 +1,274 @@
...
+
+Since we only need uncle hashes to verify the block, the number of hashes we need is at worst `log2(number-of-blocks)` and the roots of the merkle trees which has the same complexity.
+
+Notice that all new signatures verify the entire dataset since they all sign a merkle tree that spans all data. If a signed update ever conflicts against previously-verified trees, suggesting a change in the history of data, the feed is considered "corrupt" and replication will stop. This serves to disincentivize changes to old data and avoids complexity around identifying the canonical history.
+
+
+# Specifications and parameters
+[specifications-and-parameters]: #specifications-and-parameters
+
+TODO: list what kind of hashes and asymmetric keys are used
```
mafintosh: Blake2b and curve 25519

proposals/0000-hypercore.md
```diff
@@ -0,0 +1,274 @@
...
+## Private key risk
+[private-key-risk]: #private-key-risk
+
+Hypercore assumes that ownership of the private key is tantamount to authorship. If the private key is leaked, the author will lose control of the hypercore integrity. If the private key is lost, the hypercore will no longer be modifiable.
+
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+The Hypercore log is conceptually similar to Secure Scuttlebutt's log structure; both are designed to provide a single append-only history and to verify the integrity using only a public key identifier. However, Secure Scuttlebutt uses a Linked List structure with content-hash pointers to construct its log while Hypercore uses a Merkle Tree. This decision increases the amount of hashes computed and stored in Hypercore, but it enables more efficient partial replication of the dataset over the network as trees enable faster comparisons of dataset availability and verification of integrity. (Citation needed?)
```
mafintosh: Note the amount it is actually only one hash per block of data in hypercore (even numbers are data, odd ones are hashes)

proposals/0000-hypercore.md
```diff
@@ -0,0 +1,274 @@
...
+
+IPFS is designed for immutable hash-addressed content, but it provides a mechanism for mutable content using public key addresses (IPNS). IPNS is still under development but some concepts are established. Its premise is much simpler than Hypercore's; rather than encoding the history in a standard form, IPNS simply signs and publishes a content-hash identifier under the public key, therefore creating a `pubkey -> hash` lookup. The referenced content may choose to encode a history, but it is not required and no constraints on branching is enforced. Compared to Hypercore, IPNS may be more user-friendly as it does not suffer from a catastrophic error if the history splits, therefore enabling users to share private keys freely between devices. This comes at the cost that history may be freely rewritten by the dataset author.
```
mafintosh: Also means it's not a realtime change log

mafintosh: Ie in a hypercore feed you can subscribe to updates and they propagate fast and are ordered

proposals/0000-hypercore.md
```diff
@@ -0,0 +1,274 @@
...
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Is there a potential "branch resolution" protocol which could remove the [linear history requirement](#linear-history-requirement) and therefore enable users to share private keys freely between their devices? Explaining the risks of branches to users is difficult.
```
mafintosh: An extension can be made in the future that allows an author so pick a correct branch

proposals/0000-hypercore.md
emilbayes: `Blake2b-256` specifically :)

proposals/0000-hypercore.md
pfrazee: Don't you also have to deal with parent & root hashes?

mafintosh: still only one hash stored per data entry. if you include the hash of the data itself is 2x data hashes

PR #5, opened 2018-02-06
WIP: nomenclature

joehand:
This DEP is still work-in-progress.

I've added the summary/motivation and Bryan's questions. I need to put a bit of thought in how to best organize this, in the case we have a lot of terms in here.

We also have existing terms in the Dat documentation: https://docs.datproject.org/terms. I can update/consolidate those to this DEP.

### DIFF
--- a/proposals/0000-nomenclature.md
+++ b/proposals/0000-nomenclature.md
@@ -1,0 +1,63 @@
+
+Title: **DEP-0000: Dat Nomenclature**
+
+Short Name: `0000-nomenclature`
+
+Type: Informative
+
+Status: Undefined (as of 2018-02-XX)
+
+Github PR: (add HTTPS link here after PR is opened)
+
+Authors: [Joe Hand](https://github.com/joehand)
+
+
+# Summary
+[summary]: #summary
+
+The Dat Protocol is a new protocol, with new concepts and models. To ensure consistency and improve understanding, it is essential that the community agree on the definition of words and how to refer to specific concepts. This DEP aims to begin this process.
+
+# Motivation
+[motivation]: #motivation
+
+The Dat community has a variety of new terms and concepts. As these have evolved over time, there is a lack of consistency and definition of terms. This DEP will aim to document existing standard nomenclature, identifying ambiguous terms. Additionally, the DEP will aim to identify and define terms that will be preferred or required in future DEP proposals when discussing certain topics.
+
+By defining Dat nomenclature, we can ensure the writing of the wider Dat community also uses the preferred terms. This will make it easier for new users and contributors to understand the Dat ecosystem. To reduce barriers to entry, this DEP will prefer words that are less technical while conveying the same meaning. 
+
+When addressing terms that have specific technical meaning, such as *hyperdrive*, the DEP will aim to have a definition and identify other related words that should not be used in future DEPs. Terms that have less specific meaning, such as *discovery*, will have a definition and potential synonyms, but leave it to future DEPs to define any terms necessary.
+
+# Usage Documentation
+[usage-documentation]: #usage-documentation
+
+### Dat
+
+Dat refers generally to *Dat Project* or *Dat Protocol*. Dat is capitalized, not uppercase, and *Project* or *Protocol* are capitalized. 
+
+Dat can be used by itself if referring generally to a Dat-related entity, such as "the Dat community." When the meaning may be ambiguous, a DEP must include Project or Protocol.
+
+## Other Terms & Concepts
+
+
+# Rationale
+[alternatives]: #alternatives
+
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- How would it make sense to organize this?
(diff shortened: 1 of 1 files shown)


pfrazee:
Good call. I share the question about registers.

joehand:
On that topic, I started to try to add some decision making criteria when deciding between a few possible words:

>  By defining Dat nomenclature, we can ensure the writing of the wider Dat community also uses the preferred terms. This will make it easier for new users and contributors to understand the Dat ecosystem. To reduce barriers to entry, this DEP will prefer words that are less technical while conveying the same meaning. 

So, though *register* is more technically accurate it seems like *feed* or *log* may be preferred.

joehand:
Note to self from previous meeting:

> discuss syncing/seeding/etc in nomenclature DEP - jhand will formalize terms that need definition

martinheidegger:
Things I miss specified:

- `Version`: What is considered a version in DAT?
- `Bootstrapping`: What do we mean when bootstrapping the network?
- `Sparse`: DAT's can be "sparsely" replicated?!
- `Checkout`: What is considered a checkout?
- `Live`: There are properties in code that refer to something being "live"

... and a link to the terminology used in the dat protocol book: https://github.com/datprotocol/book/blob/a3ca149853b9153c7140876d6f749ecad5c6edbb/src/ch03-01-terminology.md

bnewbold:
I would define a Checkout as a folder containing files from a dat/hyperdrive feed at a specific version (which could be the most recent version; doesn't need to be "previous"). This is distinct from having the same content stored locally in SLEEP files. The terminology comes from git and `git checkout`.

To clarify Version, it's the integer message number of a hypercore feed. These days, with multi-writer, the term gets a bit more ambiguous because there are multiple feeds, so the version of a hyperdb overall can be an array of (feed, integer) pairs. UX/nomenclature around this will probably need an update for dat-on-multiwriter-hyperdb.

Sparse usually implies that not only is the dataset/feed only partially replicated, but that it's *intentionally* only partially replicated: the user only wanted, eg, a sub-directory, or only specific versions replicated. I don't think there is clarity/terminology around the case of having "the entire most recent version of a hyperdrive/hyperdb" (eg, full values for all keys/files at the most recent version) but not full history: is that considered Sparse? In conversation i've usually heard people refer to this as the default condition (just having the most recent version), and having the full history of the feed being a speciall "Full History" or "Archival" copy.

I agree with pfrazee on Bootstrapping and Live.

aral:
Suggestion regarding key naming, to strengthen the intent and usage of the keys and remove ambiguity about what abilities various keys grant:

  * Public key → Read key
  * Secret key → Write key
  * Discovery key → Discovery key (unchanged)

This way, people new to the system will not be misled into thinking, for example, that the public key is public (where would they get such an idea?) ;P

And the keys do exactly what they say on the tin.

Example usage:

> The Read Key grants read access to a DAT whereas the Write Key is required to write to a DAT. The Discovery Key is used to discover a DAT and it is derived as a hash of the Read Key. The Read Key and Write Key should both be kept secret. 

Thoughts?

yoshuawuyts:
@aral I think that's a pretty reasonable suggestion that could remove some ambiguity. I also always mistake "private key" and "secret key", which this would also help solve.

martinheidegger:
@aral I am considering working on an "encrypted DAT". :DATs that are additionally encrypted with yet another key in order to implement proxies/bridges that don't know about the content of a DAT. Do you have any idea how this Key should be called? ;)

aral:
If you mean encrypting the contents of hypercores, I’d say “encryption key” does what it says on the tin.

martinheidegger:
Thanks, naming is hard :-)

martinheidegger:
What entails "sync" in `dat sync` and "share" in `dat share`? (Question that came up in chat) Also: how to call a peer that has a write key vs  a peer that doesn't?

aral:
Here’s the latest glossary for Hypha, in case it helps – please feel free to use the definitions that apply: https://ar.al/2019/02/18/hypha-glossary/

Regarding the last question in your latest comment, @martinheidegger, authorised vs unauthorised is what I’m using.

martinheidegger:
"authorised" makes sense in a multiwriter context, but in a single-writer context (that will exist in future) it feels weird as there is no way how to every authorise another peer.

RangerMauve:
`Pinning Service`: A server you can send your `dat://` URL to in order for it to replicate your content and stay online. They're useful for making sure content is available in the network. Example: [Hasbhase](https://hashbase.io/)

martinheidegger:
@RangerMauve "pinning service" was also known as "publishing": https://github.com/datproject/dat/blob/master/src/commands/publish.js

martinheidegger:
@RangerMauve Do you think registry/publish is worse than "pinning service"/pinning?

RangerMauve:
I haven't seen `registry/publish` used outside of datbase, and I haven't seen datbased used much.

I like pinning because it's more descriptive of what's actually happening. A registry/publishing implies some sort of centralization or control. Whereas pinning has more of a "Hey, I'm keeping this around for you" feeling where you're still in control and it's no big deal who's pinning it.

I also like the term "Seeding" since it relates to the BitTorrent world

joehand:
Some of the pinning stuff may be addressed in https://www.datprotocol.com/deps/0003-http-pinning-service-api/

PR #8, opened 2018-03-04
Draft: Wire protocol (DEP)

pfrazee:
Rendered: [Draft](https://github.com/pfrazee/DEPs/blob/dep-wire-protocol/proposals/0000-wire-protocol.md)

Submitted for review. Working group will vote in next meeting (late February).


proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+
+The protocol described here is already in use as of 2017 (by Hypercore, Dat, and Beaker Browser users), and was partially described in an earlier [whitepaper][whitepaper]. This document fills in some additional details.
+
+[whitepaper]: https://github.com/datprotocol/whitepaper
+
+
+# Stream Connections
+[stream-details]: #stream-details
+
+The Dat wire protocol depends on the use of a binary transport protocol which provides the following semantics:
```
bnewbold: Should we give TCP and uTP as examples here?

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+
+
+## Channels
+[channels]: #channels
+
+Multiple hypercore feeds can be synchronized over the same protocol connection. Messages pertaining to the separate channels are tagged with an id for disambiguation.
+
+Note that at least one feed is necessary for each connection (for handshaking to succeed), and that the first feed is the one used for discovery and as an encryption key.
+
+To initiate a new channel (after the primary is established), TODO
```
bnewbold: ... send an (encrypted) Feed message followed by an Info.

It's a little unclear which party *should* initiate this (I supposed either can at any time), and what to do if both parties send a Feed message with the same index and different feed public keys (aka, how is the dispute resolved?).

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+
+Each extension is capable of sending custom payloads through the Extension message type.
+
+
+# Message Details
+[message-details]: #message-details
+
+TODO: description of framing
+
+Wire format is `<len>(<header><message>)`. `header` is a varint, of form `channel << 4 | <4-bit-type>`.
```
bnewbold: Should note that `len` includes the length of the `header` varint.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+  <tr><td> 9 <td>[Data][msg-data]
+  <tr><td>15 <td>[Extension][msg-extension]
+</table>
+
+#### Keep-Alive
+[msg-keepalive]: #msg-keepalive
+
+A message of body length 0 (giving a total message size of 1 byte for the `len` varint) is a keep-alive. Depending on transport and application needs, peers may optionally send keep-alive messages to help detect and prevent connection loss. Peers must always handle keep-alive messages correctly (aka, ignore them), regardless of transport.
+
+TODO: what is a good default interval?
```
bnewbold: From brief internet reading, sounds like surprisingly long (300 seconds / 5 minutes) works to resolve the operating system socket use-case.
I don't think it's crazy to imagine a server (like hashbase) maintaining 1000 active connections; with a 300 second interval that's already several per second.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+
+#### Feed
+[msg-feed]: #msg-feed
+
+`type=0` Should be the first message sent on a channel. Establishes the content which will be exchanged on the channel.
```
bnewbold: Would be nice to have a link back to the handshake/encryption sections earlier in the document.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+    optional bool hash = 3;
+}
+```
+
+#### Data
+[msg-data]: #msg-data
+
+`type=9` Send a Hypercore feed block. May contain the data of the block, and may contain the hashes of the ancestor nodes in the merkle tree. Should only be sent in reaction to a Request message.
+
+When a Data message is received, its node hashes and signature should be verified against any local tree information. If the nodes can be verified, they and the block data should be stored locally.
```
bnewbold: "can be" instead of "should be" here... clients might just want to cache.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+ALICE: closes connection
+```
+
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Encryption might not make sense in some contexts (eg, IPC, or if the transport layer is already providing encryption). Should this DEP recognize this explicitly?
+- What parts of the design do you expect to resolve through the DEP consensus process before this gets merged?
```
bnewbold: These template questions can probably be yanked.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+- What parts of the design do you expect to resolve through implementation and code review, or are left to independent library or application developers?
+- What related issues do you consider out of scope for this DEP that could be addressed in the future independently of the solution that comes out of this DEP?
+- How do "ack"s work?
```
bnewbold: I'm curious also!

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+These digests are very compact in size, only `(log2(number-of-blocks) + 2) / 8` bytes needed in the worst case. For example if you are sharing one trillion blocks of data the digest would be `(log2(1000000000000) + 2) / 8 ~= 6` bytes long.
+
+
+# Examples
+
+## Simple Download
+
+Alice has an e-book identified by a public-key (PK) which Bob would like to download. The e-book also has a discovery-key (DK). Bob connects to Alice, and the following exchange occurs:
+
+```
```
bnewbold: I would indicate in this table that the first two Feed messages are not encrypted.

proposals/0000-wire-protocol.md
pfrazee: Good idea

proposals/0000-wire-protocol.md
pfrazee: I'll expand on that

proposals/0000-wire-protocol.md
pfrazee: Kk

proposals/0000-wire-protocol.md
pfrazee: Ok

proposals/0000-wire-protocol.md
pfrazee: Yeah

bnewbold:
We should also write "Security and Privacy Concerns" and "References" sections for this one.

For references I could imagine:
- bittorrent, other "who are your influences?" protocol specs
- papers/references around choking and other scheduling issues (eg, see links from this [ebay post](https://www.ebayinc.com/stories/blogs/tech/bittorrent-for-package-distribution-in-the-enterprise/)
- more protobuf references (we have a single link right now)
- libsodium and hash/crypto algorithm references

For security/privacy:
- enumerate what a persistent observer can learn about peers
- enumerate some specific threat models and trust assumptions (eg, which properties of which hash and crypto algorithms are we assuming and relying on; distinction between, eg, targeted censorship and data manipulation/corruption)

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+
+The **discovery key** is generated from the public encryption key for a hypercore feed (in this case the first, or "primary" feed) by using the public key to sign the 9-byte ASCII string "HYPERCORE" (with no trailing NULL byte) using the BLAKE2b keyed hashing algorithm (provided by most BLAKE2b hash implementations). The discovery key is 32 bytes long.
+
+```js
+var discoveryKey = new Buffer(32)
+sodium.crypto_generichash(
+  discoveryKey, // out
+  'HYPERCORE',  // in (message)
+  publicKey     // key
+)
```
mafintosh: (use Buffer.alloc) also this is basically an HMAC and we should note that

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,435 @@
...
+
+#### Unhave
+[msg-unhave]: #msg-unhave
+
+`type=4` Announces the loss of availability of Hypercore feed blocks for download. The `start` and `length` represent a continuous set of blocks.
+
+This message is sent in the following contexts:
+
+ - To inform the peer that data which was sent was rejected, for instance because it was not requested and pushes are not being allowed.
+ - To announce blocks have been from local storage within a remote-wanted range.
```
andrewosh: "have been from local storage" -> "have been removed from local storage"

pfrazee:
Fixed, thanks

bnewbold:
@pfrazee what's the status on this DEP? Are we blocked waiting on anything? If you move this somewhere I have git push privs I can try to push it through in the next couple weeks.

pfrazee:
@bnewbold I think it just needs to get written, I don't think there are any blockers in dev work. I'll give you push rights on my fork.

pfrazee:
Apologies for letting it stall!

proposals/0000-wire-protocol.md
bnewbold: From grepping around hypercore/hypercore-protocol code, as far as I can tell `ack` doesn't actually do anything, other than getting passed around as an config flag.
I'm guessing that what it is intended to do is have a peer reply with a new `Have` message in response to every `Data` message it receives? @mafintosh can you clarify?

bnewbold:
Made some more progress on this, resolving most of the TODOs and comments. Still need clarification/feedback on:

- does `ack` flag actually do anything
- HMAC language

The security/privacy section in particular should get reviewed, and the whole thing probably needs a copy-edit and review for missing content (which I haven't done yet).

jIt would also be nice to have more examples, but those can take a long time to write up, and at this point, with effort being put in elsewhere (like the protocol book), I don't think it's necessary here for Draft status. There are no "rationale" or "alternatives" sections; I think those make more sense for new proposals and less for these older/existing protocol bits, but i'm open to objections.

Would be great to have this ready to vote, or at least review, by next WG meeting!

pfrazee:
Updates look good to me!

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,485 @@
...
+## Handshake Procedure
+[handshake]: #handshake
+
+A handshake procedure needs to occur for each feed on a channel; the first part of the first handshake happens in cleartext and both validates discovery keys and establishes encyption parameters used for the rest of the connection. The first (primary) channel has `id=0`.
+
+The first (cleartext) message is a Feed message, and includes two fields: a nonce and a discovery key.
+
+The **nonce** is generated by each peer as a random 32-byte sequence.
+
+The **discovery key** is generated from the public encryption key for a hypercore feed (in this case the first, or "primary" feed) by using the public key to sign the 9-byte ASCII string "HYPERCORE" (with no trailing NULL byte) using the BLAKE2b keyed hashing algorithm (provided by most BLAKE2b hash implementations). The discovery key is 32 bytes long.
```
emilbayes: It is not actually signing anything, but performing a keyed hash (as correctly described). It is a core part of the BLAKE2b algorithm, and is also called a Prefix MAC, since you essentially do `blake2b(HYPERCORE || publicKey)`. Signing would in my view involve the private key, which only the writer has access to.

proposals/0000-wire-protocol.md
emilbayes: I would personally either keep this pseudocode or actual implementation. Pseudo code would be something like `discoveryKey = blake2b(outputLength = 32, key = 'HYPERCORE', data = publicKey)`, `outputLength` being another important parameter (256 bit output)

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,485 @@
...
+
+The encryption scheme used is libsodium's stream primitive, specifically the XSalsa20 cipher. The cipher is fed a shared key (the primary hypercore feed public key), a nonce (selected by the sender and exchanged during handshake), and a block offset (representing all encrypted bytes sent on the connection in this direction so far).
+
+The specific libsodium function used is `crypto_stream_xsalsa20_xor_ic()`. Some interfacing code is necessary to process messages that don't align with the cipher's 64-byte chunk size; unused bytes in any particular chunk can be ignored. For example, if 1000 encrypted bytes had been sent on a connection already, and then a new 50 byte message needed to be encrypted and sent, then one would offset the message by `1000 % 64 = 40` bytes and XOR the first 24 bytes against block 15, then XOR the remaining 26 bytes against block 16. The bytes would be shifted back and recombined before sending, so only 50 bytes would go down the connection; the same process would be followed by the receiver.
```
emilbayes: Is this actually what happens? I know the "unofficial" overlay implementation we have in sodium-native, which we made for hypercore calculates a block at a time (64 bytes), and then buffers whatever is left over for the next invocation, making sure that no bytes of the block go unused.

This is the API I'm talking about: https://github.com/sodium-friends/sodium-native#var-instance--crypto_stream_xor_instancenonce-key-or

emilbayes:
I particularly liked the privacy/security discussion at the end. It was honest and covered all my "concerns". It's important to know that the public key is a very strong read capability, eg. that anyone with the public key can decrypt any traffic passively observed by any other peers communicating using that public key in their initial connection. So hypercores that are meant to be very sparsely replicated (eg. internet size ones) will leak a lot of metadata.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,485 @@
...
+10(1)1 <-- we already have the first uncle, 2 so don't send us that
+1(0)11 <-- we don't have the next uncle, 5
+(1)000 <-- we have the next parent, 3
+```
+
+So using this digest the recipient can easily figure out that they only need to send us one hash, 5, for us to verify block 0.
+
+The bit vector 1 (only contains a single one) means that we already have all the hashes we need so just send us the block.
+
+These digests are very compact in size, only `(log2(number-of-blocks) + 2) / 8` bytes needed in the worst case. For example if you are sharing one trillion blocks of data the digest would be `(log2(1000000000000) + 2) / 8 ~= 6` bytes long.
```
martinheidegger: Wouldn't it be 6 "hashes" long. Is one hash equal 1 byte?

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,485 @@
...
+0
+  1
+2
+    3
+4
+  5
+6
+```
+
+If we want to fetch block 0, we need to communicate whether of not we already have the uncles (2, 5) and the parent (3). This information can be compressed into very small bit vector using the following scheme:
```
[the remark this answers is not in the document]
vtduncan: I’m trying using figures to understand how this works. I still don’t grasp it, but here is where I’m at trying to draw the example shown in this section:

hash_tree_digest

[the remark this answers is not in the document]
vtduncan: My question here would be: If you just want to verify chunk 0, why not send this instead?

hash_tree_digest

proposals/0000-wire-protocol.md
bnewbold: Ah, that's smart. In my implementation I just recalculated.
I've added that the unused bytes can be saved as an optimization; in general this whole paragraph was just to clarify because I got very confused when re-implementing on my own.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,485 @@
...
+- the relative flow of content (which peer provided data to the other, and how much)
+- which primary hypercore feeds which peers have knowledge of (identified by discovery key, not public key)
+- which peers have the public key for a given discovery key (based on whether connections succeed)
+
+Such an observer can not determine the specific content of hypercore feeds, or (with confidence) which peer (or peers) have write access to a feed.
+
+The wire protocol does not pad message sizes. This means that a persistent observer could potentially identify traffic content by infering message sizes.
+
+The hypercore protocol does not intentionally identify peers across connections, but it has been shown that even the smallest amount of identifying metadata can be used, statistically, to track and surveil users. These techniques are sometimes called "device fingerprints", "browser fingerprints", or "permacookies". Hypercore protocol users should not consider themselves immune to such tracking without specific additional effort to identify and mitigate such fingerprints.
```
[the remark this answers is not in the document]
bnewbold: I've clarified this with an explicit reference to HTTP.

proposals/0000-wire-protocol.md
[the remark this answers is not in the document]
bnewbold: I rewrote this section (and will push in a minute); please let me know if it clarified things. To answer a few questions directly:

The example of wanting to verify block 0 is bad; there's a direct signature of it. I believe that when you request a block, you always get the signature for the root hash as if that was the largest block index, and you're always trying to verify that signature, so I flipped the example to block 6.

This also answers the "how do you know how high the tree is" question.

You don't need hash 1 (in the old example) because you can calculate it from hash 0 and hash 2 (you calculate hash 2 directly from the data transmitted). I guess this is also why you don't need hash 3. Note that in other situations you would need other parent hashes.

I added a reference to the Hypercore DEP which is where the parent/uncle terminology should be defined (if it's unclear there we can take a patch to improve).

A more complex (non-symmetric) example would be nice, but also maybe more confusing here, where we're really just trying to explain the binary packing, not the whole Merkle hashing scheme?

proposals/0000-wire-protocol.md
bnewbold: I moved to pseudo-code (which I also prefer to implementation or any specific language in the DEPs).

I think "keyed hash" clarifies things; maf if you want to include a note about HMAC I need something more specific.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,485 @@
...
+
+`type=1` Overall connection handshake. Should be sent just after the Feed message on the first channel only.
+
+Some notes on the fields:
+
+ - **id** Typically a 32-byte identifier which is allocated randomly by a peer at the start of its swarming session. Not authenticated. Used to help detect multiple connections to a single peer.
+ - **live** If both peers set to true, the connection will be kept open indefinitely.
+ - **userData** An open field for sending data which you can retrieve on handshake. No value is prescribed by the protocol.
+ - **extentions** A list of strings identifying additional message-types which are supported via the Extension message.
+ - **ack** Should all blocks be explicitly acknowledged? TODO how should this work?
```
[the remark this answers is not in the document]
bnewbold: Updated, thanks for the notes

bnewbold:
I think I've addressed everything above; we should give folks a few days to respond to things before a vote.

The block digest section should get review from an actual implementer (I never implemented that bit; cc: @mafintosh  @yoshuawuyts).

Could probably also use another careful proof-read, but for Draft status I think it would be good to get this out and have eyes on it.

bnewbold:
To set the timer, I will say I am submitting this for review to be voted on at the next WG meeting, though if there are significant changes needed we can push that off.
I think we discussed "should we wait for any forthcoming implementation changes/refactors in the pipe" and my feeling is to get something out that documents current released/implemented behavior, and do an update or "v2" if needed.

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,480 @@
...
+ - 3, a parent (and root hash), we do have hash
+ - root signature (of hash 3) will be sent with request
+
+Our `Request` should include the digest bits:
+
+```
+101(1) <-- indicates that the most-significant bit is a parent, not an uncle
+10(1)1 <-- do not send hash for the first uncle, 4
+1(0)11 <-- do send hash for the next uncle, 1
+(1)000 <-- do not send hash for next parent, 3
```
martinheidegger: Why is this not `(1)011` ?

martinheidegger: General remark, wouldn't this be better readable like:

```
1011 
┃┃┃┗━ indicates that the most-significant bit is a parent, not an uncle
┃┃┗━━ do not send hash for the first uncle, 4
┃┗━━━ do send hash for the next uncle, 1
┗━━━━ do not send hash for next parent, 3
```

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,480 @@
...
+```
+
+This digest (bit vector `1011`) is transmitted in a `uint64`, in the least-significant bits. The receiving peer can calculate (from the index number) exactly how many bits are expected and extract the bitfield. The "most-significant bit" referenced above is of just the fixed-size bitfield, not the `uint64` as a whole.
+
+From this, the remote peer will know to only send one hash (for block 1) for us to verify block 6. Note that we (the receiver) can calculate the hash for block 6 ourselves when we receive it.
+
+As a special case, the bit vector `1` (only contains a single one) means that the sender should not send any hashes at all.
+
+These digests are very compact in size. Only `(log2(number-of-blocks) + 2) / 8` bytes are needed in the worst case. For example if you are sharing one trillion blocks of data the digest would be `(log2(1000000000000) + 2) / 8 ~= 6` bytes long (which fits in a single `uint64`).
```
martinheidegger: And explain here how many blocks the max is:

> These digest are very compact in size, only `log2(index) + 2` bits are required. With the `uint64` (= 64 bits) number we can address [`4611686018427387904`](https://www.wolframalpha.com/input/?i=log2(4611686018427387904)+%2B+2) blocks (=the maximum length of hypercore that can be shared).

proposals/0000-wire-protocol.md
martinheidegger: Makes sense.

proposals/0000-wire-protocol.md
bnewbold: Good catch, and that formatting is much nicer, thanks!

proposals/0000-wire-protocol.md
```diff
@@ -0,0 +1,480 @@
...
+uncompressed-sequence = varint(byte-length-of-bitfield << 1 | 0) + bitfield
+```
+
+
+## Block Tree Digest
+[block-tree-digest]: #block-tree-digest
+
+As described in DEP-0002 (Hypercore), a peer should be able to verify both the integrity of received data (aka, was there corruption somewhere along the way, detected via hash) and the authenticity (aka, is this the data from the original writer, detected via signature). Hypercore transmits the hash for every data block, but only signatures for the root hashes of Merkle trees, not individual block hashes, which means a peer may need additional hashes (for data blocks they do not have a copy of) if they want to verify the signatures of individual blocks.
+
+Redundantly transmitting all such hashes on every request would be inefficient if the receiver already had some hashes, so requesting peers can specify which hashes they need in the `nodes` field of the `Request` message. Instead of sending node indexes, the peer can send a compact bitfield, indicating for each "uncle" and "parent" whether the hash should be transmitted.
```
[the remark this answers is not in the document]
bnewbold: Tweaked

proposals/0000-wire-protocol.md
bnewbold: I didn't want to use the phrasing "the maximum length of hypercore that can be shared" here, but I did put the number of blocks this can address ("This scheme works for Hypercores with up to `2^62 = 4,611,686,018,427,387,904` blocks.")

bnewbold:
Thanks for review anybody!

urbien:
Was this Draft promoted to the Current (or however it is called) status?

PR #11, opened 2018-03-18
For Discussion: timestamps in hyperdb

bnewbold:
[Rendered pre-merge](https://github.com/bnewbold/dat-deps/blob/dep-hyperdb-timestamps/proposals/0000-hyperdb-timestamps.md)

Proposes including "event" timestamps at the hyperdb abstraction layer for every change to the database.

I suspect @mafintosh would prefer doing this at the highest abstraction level as possible (eg, at hyperdrive, not hyperdb or hypercore), but i'm curious what others think, and whether this would be valuable at all.

Before seriously reviewing to merge as Draft, I would want to demonstrate working code and have a better idea what the API would look like, but this is otherwise pretty flushed out.

### DIFF
--- a/proposals/0000-hyperdb-timestamps.md
+++ b/proposals/0000-hyperdb-timestamps.md
@@ -1,0 +1,153 @@
+
+Title: **DEP-0000: Hyperdb Timestamps**
+
+Short Name: `0000-hyperdb-timestamps`
+
+Type: Standard
+
+Status: Undefined (as of YYYY-MM-DD)
+
+Github PR: (add HTTPS link here after PR is opened)
+
+Authors: [Bryan Newbold](https://github.com/bnewbold)
+
+
+# Summary
+[summary]: #summary
+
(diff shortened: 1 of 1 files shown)


pfrazee:
I dont have much to say except I'm +1 the premise and my gut reaction is that hyperdb is the right place for this (not hypercore).

bnewbold:
This got discussed here: https://github.com/mafintosh/hyperdb/issues/97#issuecomment-387048934

With the new hyperdb `deleted` flag, even a deletion node can contain a value. This means that a higher-level application can include a timestamp in every hyperdb value/node.

I still think a timestamp might be "special" enough of a form of metadata to be included in the hyperdb Entry protobuf schema as an optional field, but I don't feel particularly urgent about it. Going to let this DEP PR linger, but wouldn't oppose closing it if we do a garbage collection at some point.

martinheidegger:
Referencing: https://github.com/mafintosh/append-tree/issues/16

PR #16, opened 2018-03-23
Clarify the maximum number of roots sentence.

bcomnes:
This sentence is slightly unclear, and slightly contradictory to the statement preceding it.  Adding a few extra words of clarification improves its intention.

### DIFF
--- a/proposals/0002-hypercore.md
+++ b/proposals/0002-hypercore.md
@@ -171,7 +171,7 @@ The nodes in this tree would be calculated as follows:
 10 = h(chunk5)
 ```
 
-In the Hypercore feed, we only want one active root. Therefore, when there are multiple roots we hash all the roots together again. At most there will be `log2(number of data blocks)`.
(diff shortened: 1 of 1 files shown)


pfrazee:
@bnewbold yeah I think that's a good improvement

bcomnes:
Agreed, justifying why we want a single fixed size hash is +1.

"To encode in the Hypercore feed" Is ambiguous.  Encoded in the merkle tree forever? Encoded by convention until there are a different set of roots?

bcomnes:
"We always want a way to refer to a single fixed-size hash of the hypercore feed.  Therefore, when there are multiple roots, we re-hash all the individual root hashes together into a single hash value. At most there will be log2(number of data blocks) such individual root hashes to re-hash."

Something like that, plus clarification on where that fixed size hash is stored / encoded?  Is my understanding correct in that this hash doesn't ever end up in the feed?

bnewbold:
Merged PR #29, which was a variant on this.

bnewbold:
Thank you @bcomnes!

PR #19, opened 2018-04-01
Draft: HTTP Pinning Service API DEP

pfrazee:
Implementations:

 - Server: https://github.com/beakerbrowser/homebase
 - Client: https://github.com/beakerbrowser/dat-pinning-service-client

### DIFF
--- a/proposals/0000-http-pinning-service-api.md
+++ b/proposals/0000-http-pinning-service-api.md
@@ -12,12 +12,14 @@ Github PR: (add HTTPS link here after PR is opened)
 Authors: Paul Frazee
 
 
-# Summary [summary]: #summary
+# Summary [summary]:
+#summary
 
 An HTTP API for adding and removing Dat data.
 
 
-# Motivation [motivation]: #motivation
+# Motivation
+[motivation]: #motivation
 
 Users frequently make use of "pinning services" to keep their Dat data online
 independently of their personal devices. By specifying a standard API for
@@ -30,7 +32,8 @@ dat publish --name myarchive my-pinning-service.com
 ```
 
 
-# Service description (PSA) document [service-description]: #service-description
+# Service description (PSA) document
+[service-description]: #service-description
 
 Servers should host the PSA service-description document at `/.well-known/psa`.
 It may be fetched using a GET request. This document will fit the following schema:
@@ -63,7 +66,8 @@ with the final URLs for their specifications.) If either API is absent from
 the PSA document, the service will be rejected.
 
 
-# User accounts API [user-accounts-api]: #user-accounts-api
+# User accounts API
+[user-accounts-api]: #user-accounts-api
 
 The user-accounts API should provide the following resources:
 
@@ -86,7 +90,7 @@ Full documentation for this API should be made available at https://user-account
 
 User POSTS to `/register` with body:
 
-```json
+```
 {
   email: String
   username: String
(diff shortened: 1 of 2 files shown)


proposals/0000-http-pinning-service-api.md
```diff
@@ -0,0 +1,216 @@
...
+The dat pinning API should provide the following resources:
+
+```
+GET  /            List all Dat data pinned by this account.
+POST /add         Add a Dat to this account's list of pins.
+POST /remove      Remove a Dat from this account's list of pins.
+GET  /item/:key   Get information about a Dat in the account's list of pins.
+                  Key may be the pubkey or name of the dat.
+POST /item/:key   Update information about a Dat in the account's list of pins.
+                  Key may be the pubkey or name of the dat.
```
mafintosh: do we need to discuss the payloads of the above or just make it up as we go along?

pfrazee:
I plan to specify that and each route in detail

pfrazee:
Working implementation is available at https://github.com/beakerbrowser/homebase

pfrazee:
I put together a client module as well, no dependencies, runs automated tests against homebase: https://github.com/beakerbrowser/dat-pinning-service-client

RangerMauve:
Is there a requirement for these services to have CORS headers so that they could be interacted with from within a webpage?

pfrazee:
Would we want CORS to be enabled?

RangerMauve:
If CORS isn't set to `*` you won't be able to do POSTs in JS AFAIK.

Not sure if there were _more_ caveats for the `dat://` origin.

pfrazee:
Yeah I'm not sure we want to enable web pages to do that yet. Users would have to give their username/password to the site to do the login flow. For Beaker, I suspect we'd want to create some kind of utility for managing that access.

RangerMauve:
I originally wrote a big rant, but the summary is that I think that nudging people to provide CORS headers for all PSAs will enable more interesting applications. Even if you don't want to explicitly require it now, this can be changed in the future and could be enabled by hosts that want to allow it.

<details>
<summary>My original rant</summary>
I get the use case for beaker, but I could envision people making apps for doing similar things that could work with these services regardless of browser. If Beaker provides a great way to integrate with these services, people are less likely to go apps that implement the same thing.

I could however see myself making a private app so I can manage my dats from either Beaker, Bunsen (before they integrate it), or the upcoming browser extensions. Locking these services to make them only usable to the browser runtime itself will make it harder to innovate.

I agree that a pinning service that takes credentials might want to be more guarded against letting web apps use it, but I could see pinning services without credentials, or completely different (e.g. discovery) services that would be more useful if apps could do whatever they wanted with them.

Explicitly specifying that only browsers will be using the DEPs will set a precedent that will stifle innovation and could lead to PSAs being used by beaker and not by third parties.
</details>

pfrazee:
That's a fair point.

bnewbold:
This looks good to me for Draft status

millette:
Did you mean to link to [archive.org](http://web.archive.org/web/20180418164539/https://github.com/beakerbrowser/beaker/wiki/PSA-Web-Service-Discovery-Protocol) instead of [direct a link](https://github.com/beakerbrowser/beaker/wiki/PSA-Web-Service-Discovery-Protocol)? It appears twice.

pfrazee:
@millette yes, that was intentional to make sure the linked target doesnt change

PR #25, opened 2018-04-27
Add DNS DEP

pfrazee:
Ready for review!

### DIFF
--- a/proposals/0000-dns.md
+++ b/proposals/0000-dns.md
@@ -1,0 +1,124 @@
+
+Title: **DEP-0000: DNS**
+
+Short Name: `0000-dns`
+
+Type: Informative
+
+Status: Draft (as of 2018-04-27)
+
+Github PR: (add HTTPS link here after PR is opened)
+
+Authors: Paul Frazee
+
+
+# Summary
+[summary]: #summary
+
+Dat's data structures (HyperCore, HyperDB, and HyperDrive) are addressed using
+cryptographic keys. In the context of Web browsers, a URL scheme is used which
+is structured as `'dat://' {key} '/' {path...}`.
+
+This DEP describes an additional protocol Dats using DNS short names.
+
+
+# Motivation
+[motivation]: #motivation
+
+The cryptographic keys which address Dats are secure and global, but not human-readable. This creates a usability problem.
+
+The goal of this DEP is to leverage DNS to provide human-readable shortnames which map to Dat's cryptographic addresses. The solution...
+
+ - Must provide a single canonical Dat URL for the domain. It should not be possible for a name to have multiple valid mappings. (No conflict.)
+ - Must not be controllable by non-owners of the domain. It should not be possible for third parties to modify the mapping. (Secure.)
+ - Should be accessible to as many users as possible. (Convenient.)
+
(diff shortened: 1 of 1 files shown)


pfrazee:
Do we want to think about a procedure for [DNS wildcards](https://en.wikipedia.org/wiki/Wildcard_DNS_record)? Something where, when that matches, you open the wildcard dat and then lookup the subdomain record in its content?

proposals/0000-dns.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+# Resolution process
+[resolution-process]: #resolution-process
+
+Resolution of a Dat at `dat://{name}` should follow this process:
+
+ - Client checks its names cache. If a non-expired entry is found, return with the entry.
+ - Client issues a DNS TXT request for `name`. This request should be issued via a secure transport (see ["DNS-over-HTTPS"](#dns-over-https)).
+ - Client iterates all TXT records given (skip if none). If a record's value matches the RegExp `^dat:\/\/[0-9a-f]{64}\/?$`:
```
mafintosh: The TXT record is a series of key=values itself. We should pick a key here here that contains the dat key.

proposals/0000-dns.md
```diff
@@ -0,0 +1,124 @@
...
+   - If the record includes a non-zero TTL, store the entry in the names cache.
+   - Return the entry.
+ - Client issues an HTTPS GET request to `https://{name}/.well-known/dat`.
+   - If the server responds with a `404 Not Found` status, client stores a `null` entry in the names cache with a TTL of 3600 and returns a failed lookup.
+   - If the server responds with anything other than a `200 OK` status, return a failed lookup.
```
bnewbold: To be explicit, this means no retries or following of redirects (3xx)? This makes the scheme more brittle than normal HTTP behavior. As a concrete example, all archive.org downloads redirect from `archive.org/download/some-item/some-file.txt` to the specific storage server using an HTTP redirect. `.well-known` might not seem like the sort of thing that would need redirects to us today, but it might be something folks want or expect or need in the future.
I'm not entirely sure what the right thing to do here is; allowing redirects could make reasoning about behavior harder.

proposals/0000-dns.md
```diff
@@ -0,0 +1,124 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+The cryptographic keys which address Dats are secure and global, but not human-readable. This creates a usability problem.
+
+The goal of this DEP is to leverage DNS to provide human-readable shortnames which map to Dat's cryptographic addresses. The solution...
+
+ - Must provide a single canonical Dat URL for the domain. It should not be possible for a name to have multiple valid mappings. (No conflict.)
```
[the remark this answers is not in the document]
pfrazee: The point about "No conflict" is that there should be a well-defined solution for getting a single global mapping. I'm not sure I consider cache-freshness to be a factor there: that's more a question of consistency strictness.

I'm not sure geo-targeted dats should be handled at the names layer but it's an interesting idea.

pfrazee:
Added a "Security and Privacy" section, and rewrote the DNS TXT record to use the following schema:

```
DATKEY={key}
```

I chose that keyname because the existing DNS record types [do a similar app-specific prefix](https://en.wikipedia.org/wiki/List_of_DNS_record_types) (IPSECKEY, OPENPGPKEY) but I'm open to discussion.

RangerMauve:
I like the `DATKEY=` prefix. IPNS uses something similar with `dnslink=`

PR #28, opened 2018-06-06
DEP: Ephemeral message extension

pfrazee:
Another DEP proposal

> This DEP defines the non-standard `em` extension message used in the Dat replication protocol. This message provides a way to send arbitrary application data to a peer through an existing connection.

This is not meant to supersede https://github.com/datprotocol/DEPs/pull/27 and will probably be used in combination with it.

### DIFF
--- a/proposals/0000-ephemeral-message.md
+++ b/proposals/0000-ephemeral-message.md
@@ -7,7 +7,7 @@ Type: Informative
 
 Status: Undefined (as of 2018-06-05)
 
-Github PR: 
+Github PR: https://github.com/datprotocol/DEPs/pull/28
 
 Authors: [Paul Frazee](https://github.com/pfrazee)
 
@@ -51,5 +51,5 @@ After publishing this DEP, the "Beaker Browser" will implement a Web API for exp
 # Changelog
 [changelog]: #changelog
 
-- 2018-05-31: First complete draft submitted for review
+- 2018-06-05: First complete draft submitted for review
 
--- a/proposals/0000-ephemeral-message.md
+++ b/proposals/0000-ephemeral-message.md
@@ -44,8 +44,8 @@ After publishing this DEP, the "Beaker Browser" will implement a Web API for exp
 [drawbacks]: #drawbacks
 
 - This DEP may present privacy concerns, as it may be used to track users in a similar fashion to HTTP Cookies, or be used to exfiltrate data.
-- The payload of the `'session-data'` message is not authenticated in any way. The lack of trust must be considered by applications which leverage the data.
-- If the recipient of the `'session-data'` message is not authenticated (as is currently the case in all Dat replication connections) the client will not know who is receiving the payload and may broadcast sensitive information.
+- The payload of the `'em'` message is not authenticated in any way. The lack of trust must be considered by applications which leverage the data.
+- If the recipient of the `'em'` message is not authenticated (as is currently the case in all Dat replication connections) the client will not know who is receiving the payload and may broadcast sensitive information.
 
 
 # Changelog
--- a/proposals/0000-ephemeral-message.md
+++ b/proposals/0000-ephemeral-message.md
@@ -1,0 +1,55 @@
+
+Title: **DEP-0000: Ephemeral Message (Extension Message)**
+
+Short Name: `0000-ephemeral-message-extension`
+
+Type: Informative
+
+Status: Undefined (as of 2018-06-05)
+
+Github PR: 
+
+Authors: [Paul Frazee](https://github.com/pfrazee)
+
+
+# Summary
+[summary]: #summary
+
+This DEP defines the non-standard `em` extension message used in the Dat replication protocol. This message provides a way to send arbitrary application data to a peer through an existing connection.
+
+
+# Motivation
+[motivation]: #motivation
+
+While Dat is effective at sharing persistent datasets, applications frequently need to transmit extra information which does not need to persist. This kind of information is known as "ephemeral." Examples include: sending chat messages, proposing changes to a dat, alerting peers to events, broadcasting identity information, and sharing the URLs of related datasets.
+
+This extension message will establish a common mechanism for sending ephemeral messages of an arbitrary composition.
+
+
+# Reference Documentation
+[reference-documentation]: #reference-documentation
+
+This DEP is implemented using the Dat replication protocol's "extension messages." In order to broadcast support for this DEP, a client should declare the `'em'` extension in the replication handshake.
+
+Ephemeral messages can be sent at any time after the connection is established by sending an extension message of type `'em'`. The message may include a payload up to 256 bytes in length. Any additional bytes should be truncated by the receiving client. The payload is a buffer of any encoding.
+
+The client may respond to the message by emitting an event, so that it may be handled by the client's application logic.
+
+No acknowledgment of receipt will be provided (no "ACK").
+
+After publishing this DEP, the "Beaker Browser" will implement a Web API for exposing the `'em'` protocol to applications. It will restrict access so that the application code of a `dat://` site will only be able to send ephemeral messages on connections related to its own content.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- This DEP may present privacy concerns, as it may be used to track users in a similar fashion to HTTP Cookies, or be used to exfiltrate data.
+- The payload of the `'session-data'` message is not authenticated in any way. The lack of trust must be considered by applications which leverage the data.
+- If the recipient of the `'session-data'` message is not authenticated (as is currently the case in all Dat replication connections) the client will not know who is receiving the payload and may broadcast sensitive information.
+
+
+# Changelog
+[changelog]: #changelog
+
+- 2018-05-31: First complete draft submitted for review
+
--- a/proposals/0000-ephemeral-message.md
+++ b/proposals/0000-ephemeral-message.md
@@ -31,9 +31,20 @@ This extension message will establish a common mechanism for sending ephemeral m
 
 This DEP is implemented using the Dat replication protocol's "extension messages." In order to broadcast support for this DEP, a client should declare the `'em'` extension in the replication handshake.
 
-Ephemeral messages can be sent at any time after the connection is established by sending an extension message of type `'em'`. The message may include a payload up to 256 bytes in length. Any additional bytes should be truncated by the receiving client. The payload is a buffer of any encoding.
+Ephemeral messages can be sent at any time after the connection is established by sending an extension message of type `'em'`. The message includes 1 byte of header information followed by a payload no longer than 10 kilobytes. Any additional bytes should be truncated by the receiving client. Therefore the message is encoded as follows:
 
-The client may respond to the message by emitting an event, so that it may be handled by the client's application logic.
+```
+header  (1 byte)
+payload (0-10000 bytes)
+```
+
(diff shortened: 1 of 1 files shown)


pfrazee:
Here's a lab API I'm considering for Beaker to leverage both this and #27 https://gist.github.com/pfrazee/7259f9201d44417c777803984715e1c4

pfrazee:
> Why 1000 bytes?

I originally had it at 256 bytes to fit within a single packet (with some clearance) but I decided that's just too much of a pain for users.

I figure we should cap it at *some* size to discourage DoS attacks (ie "hey parse this 100mb json message!"). I'm open to discussing the exact size, but 1kb seemed reasonable.

> What happens when there's a parsing error for the payload?

I figure you just drop the message. There's no acknowledgement of receipt and so an error won't come back either.

> Are the peer IDs going to be random per dat site?

They're random per connection, at the moment.

RangerMauve:
Also, question: In the future would it be weird if I did the "approve" review action when reviewing stuff even though I'm not part of the WG? I almost did it by reflex but wasn't sure if it would be weird to do so.

pfrazee:
Straight into the dungeon with you

proposals/0000-ephemeral-message.md
```diff
@@ -0,0 +1,67 @@
...
+# Summary
+[summary]: #summary
+
+This DEP defines the non-standard `em` extension message used in the Dat replication protocol. This message provides a way to send arbitrary application data to a peer through an existing connection.
+
+
+# Motivation
+[motivation]: #motivation
+
+While Dat is effective at sharing persistent datasets, applications frequently need to transmit extra information which does not need to persist. This kind of information is known as "ephemeral." Examples include: sending chat messages, proposing changes to a dat, alerting peers to events, broadcasting identity information, and sharing the URLs of related datasets.
```
bnewbold: Instead of adding general purpose capabilities to the Dat protocol to support all these use cases, why not embed Dat withing application-specific protocols for each of these use cases? The protocol stack is pretty modular as-is, so it should be possible to borrow "just" features like discovery, stream encryption, etc. Hypercore is transport agnostic, so it should be possible to embed in each of these.

This is sort of a rhetorical/hypothetical question; what i'm really getting at is, are we trying to turn the hypercore protocol a general purpose distributed networking framework? If so, maybe we should draw up a roadmap of what that would entail.

proposals/0000-ephemeral-message.md
```diff
@@ -0,0 +1,67 @@
...
+
+This extension message will establish a common mechanism for sending ephemeral messages of an arbitrary composition.
+
+
+# Reference Documentation
+[reference-documentation]: #reference-documentation
+
+This DEP is implemented using the Dat replication protocol's "extension messages." In order to broadcast support for this DEP, a client should declare the `'em'` extension in the replication handshake.
```
bnewbold: Does `em` stand for "extension message" or "ephemeral message"? I would expand it to `ephemeral-message`, or `ephemeral` if we're trying to save bytes.

proposals/0000-ephemeral-message.md
```diff
@@ -0,0 +1,67 @@
...
+Ephemeral messages can be sent at any time after the connection is established by sending an extension message of type `'em'`. The message includes 1 byte of header information followed by a payload no longer than 10 kilobytes. Any additional bytes should be truncated by the receiving client. Therefore the message is encoded as follows:
+
+```
+header  (1 byte)
+payload (0-10000 bytes)
+```
+
+The first 6 header bits are reserved. The payload's encoding may be specified in the last two header bits. Possible values are:
+
+ - `00` - binary
```
bnewbold: Hrm, this is pretty limited. I think it might be better to just not attempt to do content-type declaration at all rather than do it partially. An HTTP-style Content-Type header feels like the minimum viable here. How will the client know how to "decode the payload according to this encoding-value" at all if it doesn't know what the message type is (only the encoding)?

proposals/0000-ephemeral-message.md
```diff
@@ -0,0 +1,67 @@
...
+ - `01` - utf8
+ - `10` - json
+
+The receiving client should attempt to decode the payload according to this encoding-value. The client may respond to the message by emitting an event, so that it may be handled by the client's application logic.
+
+No acknowledgment of receipt will be provided (no "ACK").
+
+After publishing this DEP, the "Beaker Browser" will implement a Web API for exposing the `'em'` protocol to applications. It will restrict access so that the application code of a `dat://` site will only be able to send ephemeral messages on connections related to its own content.
```
bnewbold: I'd leave forward-looking statements ("this client will support XYZ") out of the DEP, even at draft stage. The note about expected security policy might be worth keeping though.

proposals/0000-ephemeral-message.md
pfrazee: Beaker needs a quick solution to exchanging ephemeral messages between users that reside on a dat:// site. Hypercore has an extensions mechanism for sending arbitrary message types, and so this DEP is taking advantage of that. It's a low-effort, high-return feature; just a stepping stone to the broader roadmap.

proposals/0000-ephemeral-message.md
pfrazee: I like `ephemeral`

proposals/0000-ephemeral-message.md
pfrazee: I'm just looking for a way to send buffers, strings, or objects. Maf had some observations about using a protocol buffer so that we can extend it more in the future, which I agree with.

proposals/0000-ephemeral-message.md
pfrazee: What's a better way to phrase it, do you think? Maybe as a recommendation to browser clients which implement the DEP?

bnewbold: I would move it to "Motivation" and rephrase:

"A specific use case for this extension is to enable a new Web API which will expose peer message-passing channels to in-browser applications. Such an API would restrict access so that the application code of a `dat://` site will only be able to send ephemeral messages on connections related to its own content."

proposals/0000-ephemeral-message.md
pfrazee: Makes sense to me. Should we use the full mime-type? `"application/json"` ?

proposals/0000-ephemeral-message.md
pfrazee: Yeah that's good

pfrazee:
Changes made. I removed any max payload length. Do we want to readd that?

proposals/0000-ephemeral-message.md
bnewbold: Yup; and applications can support later variants if they want (eg, for GeoJSON, `application/geo+json`)

bnewbold:
Could go either way on payload length... if we do limit it i'd leave the door open to increasing the size in the future, but never decreasing the message size.

proposals/0000-ephemeral-message.md
RangerMauve: I really like the idea of using mime types! It'll make packets a bit bigger, but I think it's worth the flexability it will bring.

pfrazee:
Is this 👍 to publish as a draft?

bnewbold:
It would be nice to at least mention that payload size should be kept "reasonable", but I don't have specific language to recommend. I wouldn't block on that.

Skimming back over this, I don't see a privacy/security section. Even as an "Informative" and "Draft", I think there should be a section, and at least a "user/developer beware" statement. This in addition to the statement under drawbacks. Off the top of my head:

- messages are encrypted on the wire and have the same global observer privacy properties as regular hypercore messages in that regard (eg, "what can your gateway router or ISP inspect/filter/block").
- the discovery key semantics mean that you know you are connecting to a peer that knows the discovery key. however, within that set of users (which could be large/public/everybody if the discovery key is a public app), there is no man-in-the-middle protection; any protocol trying to implement authentication or secure message using ephemeral messages needs to treat them like any other datagram over the open internet; hypercore does not provide a trustworthy end-to-end or peer-to-peer messaging channel. as with hypercore, the current status of the network is that a gateway can "trivially" control discovery of peers (unless DNS-over-TLS or DNSSEC or similar), and as well anybody can join the swarm, so man-in-the-middling is a real thing for all users, not an advanced-persistent-threat scenario; this isn't as much of a concern with regular hypercore because you're verifying the data you receive by signature, but ephemeral messages aren't signed with the feed key (unless the application developer does so themselves)

That last point kind of ran on, sorry about that.

It would be nice to link to the implementation of extension messages, but we don't have the wire protocol DEP published yet. For reference, the message type is 15 (`0x0F`). Also the semantics: if ephemeral messages are being used, no other extension messages can be sent unambiguously, and the "scope" of extension messages is to the connection, not to any particular feed or channel (though I guess it's implicitly tied to the first/primary feed that was used to initiate the connection).

Of the above, the only thing i'd consider blocking is an additional security/privacy section. Apologies for not noticing and requesting earlier; I can draft something if you want.

pfrazee:
@bnewbold I'm happy to write it, and happy for you to! I'd probably steal what you wrote their either way. Just LMK

bnewbold:
If you're up for writing it please do, and feel free to take anything from the above.

pfrazee:
@bnewbold glad you suggested that section, it was needed. Ready for review.

bnewbold:
We should discuss in the WG meeting, but after thinking about it more I think the above security issue is serious enough that it feels like we're handing developers a foot-gun by publishing this as a DEP, even a Draft. This doesn't feel "safe by default": there is a large burden on application developers and users to understand the security/privacy semantics.

pfrazee:
We decided to withdraw this proposal. The WG was uncomfortable giving this spec approval given its security & privacy characteristics. Applications are (as always) free to use similar designs in their own extension messages.

RangerMauve:
Is the session data proposal still good to go?

pfrazee:
@bnewbold do you have any feelings about that? It's not fundamentally different in its security & privacy properties.

RangerMauve:
Also, if this is being revoked, does that mean the datPeers API in beaker is going away? Would it be possible to keep something like it for sessionData if it's just ephemeral messages that are going away?

pfrazee:
No the `datPeer` API is a go as planned. We decided not to make it an official DEP

aral:
Sorry to revive a closed PR but regarding the privacy concerns, wouldn’t they be addressed by having the ephemeral messages encrypted by the feed key as part of the ephemeral extension protocol itself (instead of leaving it to userland?)

It feels like this would be preferable to every userland application creating its own method of handling ephemeral messaging.

pfrazee:
@aral AFAIK that is the case -- all messages over the `hypercore-protocol` are encrypted using the feed key that arranged the connection. The concern is that the feed key is not secret enough for general use. For instance, if the communication were occurring over a popular app's dat.

There's a lot of work planned & under way with the discovery and connection layer. Discovery is shifting to [hyperswarm](https://github.com/hyperswarm) and connection-layer encryption is (last I heard) moving to a [NOISE](http://www.noiseprotocol.org/) implementation. The reasons for this include:

 - Enabling authenticated dat connections, so that you can specify a network ACL (eg "only sync with aral").
 - Hyperswarm moves discovery onto a DHT, which improves resilience. Hopefully that work will improve overall connection-reliability as well.
 - We have plans to make the discovery+connection layer a separate reusable stack which Beaker will expose as the [Peer Socket API](https://github.com/beakerbrowser/beaker-core/pull/6). This will make it possible to arrange connections without piggybacking on the replication streams and that should make the connections fit their intended uses a little better in terms of controlling who participates in them.

PR #33, opened 2018-06-26
Remove hyperdrive-specific material from the hypercore spec

pfrazee:
I noticed that we have a hyperdrive-specific description of using a `metadata` and `content` feed in the intro of this spec. I think that's somewhat misleading (especially because it characterizes "Dat" as using these two feeds). I suggest we simply remove it, because the spec is fine without it.


bnewbold:
I think it would be good to leave a mention (sort of a learning and context pointer) that higher-level formats make use of multiple hypercores, not just one. I know when I was originally learning how dat worked this was unexpected.

PR #34, opened 2018-06-26
Hypercore Header DEP

pfrazee:
Standardizes a `HypercoreHeader` message which can be used as the first entry in hypercores to declare the datastructure along with custom fields.

### DIFF
--- a/proposals/0000-hypercore-header.md
+++ b/proposals/0000-hypercore-header.md
@@ -23,7 +23,7 @@ The Dat protocol supports a variety of data structures on top of the hypercore a
 
 At time of writing, Dat's standard data structures are being expanded upon (hyperdb) while existing data structures are being updated with breaking changes (hyperdrive). In order for these changes to be deployed smoothly to the network, it's important that the following needs be met:
 
- 1. Programs should correctly identify the data structure type and version within a hypercore. This will enable the program to use the correct handling code (eg hyperdb, hyperdrive-v1, hyperdrive-v2).
+ 1. Programs should correctly identify the data structure type and version within a hypercore. This will enable the program to use the correct handling code (eg hyperdb, hyperdrive, hyperdrive-v2).
  2. Programs should abort processing a hypercore if they lack support for the type or version of the data structure.
 
 
--- a/proposals/0000-hypercore-header.md
+++ b/proposals/0000-hypercore-header.md
@@ -30,9 +30,9 @@ At time of writing, Dat's standard data structures are being expanded upon (hype
 # Usage Documentation
 [usage-documentation]: #usage-documentation
 
-The "header" is the first entry in a hypercore. It includes a `type` and an optional `extension` which can contain data-structure-specific data.
+The "header" is the first entry in a hypercore. It includes a `dataStructureType` and an optional `extension` which can contain data-structure-specific data.
 
-A program that is reading a hypercore will examine the `type` to determine how to process the hypercore.
+A program that is reading a hypercore will examine the `dataStructureType` to determine how to process the hypercore.
 
 ```
 function loadCorrectStructure (hypercore) {
@@ -44,7 +44,7 @@ function loadCorrectStructure (hypercore) {
     return hypercore
   }
 
-  switch (header.type) {
+  switch (header.dataStructureType) {
     case 'hyperdrive':
       return new HyperdriveV1(hypercore, header)
     case 'hyperdrive-v2':
@@ -60,7 +60,7 @@ function loadCorrectStructure (hypercore) {
 }
 ```
 
-The `type` string should be unique to the data structure (refer to existing DEPs to avoid conflicts). When a breaking change is made to the data-structure, it should be given a new `type` string. For example, the `"hyperdrive"` type string will be updated to `"hyperdrive-v2"` for its second version.
+The `dataStructureType` string should be unique to the data structure (refer to existing DEPs to avoid conflicts). When a breaking change is made to the data-structure, it should be given a new `dataStructureType` string. For example, the `"hyperdrive"` type string might be updated to `"hyperdrive-v2"` for its second version. (Note: the actual type strings will be specified in the DEPs for the individual data structures.)
 
 The `header.extension` is an optional blob of bytes. It can be used by the data structure to store custom data. In `"hyperdrive"` (v1) it is used to specify the key of the content hypercore. It is also valid to encode a protobuf message into the extension field, and therefore it's simple to add additional fields.
 
@@ -74,7 +74,7 @@ The header message has the following protobuf schema:
 
 ```protobuf
 message HypercoreHeader {
-  required string type = 1;
+  required string dataStructureType = 1;
   optional bytes extension = 2;
 }
 ```
@@ -87,7 +87,7 @@ message MyCustomHeader {
     required string foo = 1;
     required uint64 bar = 2;
   }
-  required string type = 1;
+  required string dataStructureType = 1;
   optional AdditionalData extension = 2;
 }
 ```
--- a/proposals/0000-hypercore-header.md
+++ b/proposals/0000-hypercore-header.md
@@ -79,7 +79,7 @@ message HypercoreHeader {
 }
 ```
 
-Data structures should not add additional fields to the `HypercoreHeader`. They should put any additional data into the `extension`. This is easy to do with protobuf's nested messages. For example:
+Data structures should not add additional fields to the `HypercoreHeader`. Any additional fields should be ignored. Data-structure-specific data can be stored in the `extension`. This is easy to do with protobuf's nested messages. For example:
 
 ```protobuf
 message MyCustomHeader {
--- a/proposals/0000-hypercore-header.md
+++ b/proposals/0000-hypercore-header.md
@@ -34,30 +34,29 @@ The "header" is the first entry in a hypercore. It includes a `type` and an opti
 
 A program that is reading a hypercore will examine the `type` to determine how to process the hypercore.
 
-```js
-function loadCorrectStructure (hypercore, cb) {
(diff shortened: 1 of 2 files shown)


proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,109 @@
...
+  required string type = 1;
+  optional AdditionalData extension = 2;
+}
+```
+
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+This standard is backwards compatible with v1 of hyperdrive, which used an `Index` message as the first entry in the metadata hypercore.
```
bnewbold: Oh, :party:!

proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,109 @@
...
+The header message has the following protobuf schema:
+
+```protobuf
+message HypercoreHeader {
+  required string type = 1;
+  optional bytes extension = 2;
+}
+```
+
+Data structures should not add additional fields to the `HypercoreHeader`. They should put any additional data into the `extension`. This is easy to do with protobuf's nested messages. For example:
```
bnewbold: Should implementations fail fast on additional fields, or should they ignore additional fields? The later is future-compatible, the former is not.

proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,109 @@
...
+        return cb(null, hypercore) // no structure, treat as a hypercore
+      // ...
+      default:
+        return cb(new Error('Unsupported: ' + header.type))
+    }
+  })
+}
+```
+
+The `type` string should be unique to the data structure (refer to existing DEPs to avoid conflicts). When a breaking change is made to the data-structure, it should be given a new `type` string. For example, the `"hyperdrive"` type string will be updated to `"hyperdrive-v2"` for its second version.
```
bnewbold: Does this imply that the `-` character is special or reserved? Should successor versions use the `-vNNN` syntax?
To clarify my bikeshedding, I think encoding the version as an unstructured string is a reasonable choice, but I think a default delimiter between "schema" and "version/variant" should be chosen. 

Some common aesthetically pleasing separators would be `-`, `/`, `.`, `_`. How do you read `one-two_three-four`? As two identifiers (`one-two` followed by `three-four`), or three (`one`, `two_three`, `four`)?

To propose something concrete, lets say dash (`-`) is the separator, and identifiers should have dashes in them. Let's also say that the version is optional but recommended ("hyperdrive-v0"), and that the `x-` or `X-` prefixes are interpreted as "experimental" (aka, no DEP exists yet).

proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,109 @@
...
+
+
+# Usage Documentation
+[usage-documentation]: #usage-documentation
+
+The "header" is the first entry in a hypercore. It includes a `type` and an optional `extension` which can contain data-structure-specific data.
+
+A program that is reading a hypercore will examine the `type` to determine how to process the hypercore.
+
+```js
```
bnewbold: I have a soft preference for using pseudo code rather than literal javascript in the DEPs. In additional to being more implementation-agnostic, an argument against using actual code is that it comes with an implication that this exact API is stable, or that this exact code is the canonical code that people should always copy/paste into their applications even years down the line.

pfrazee:
> I still kind of wish we could deal with nested schemas

I'm not against revisiting this. It should pretty much always be layered (one structure on top of another) so all we need is a standard separator. The downside is that it adds some complexity (multiple types emitted instead of just 1).

>  I don't remember that much about the wire representation of protobuf, but we
will implicitly still be trying to parse a lot of random bytestrings to see if
they "match" this schema.

Hmm, yeah that's true.

> Should we mention that folks should specifically avoid having a protobuf message that has a string field as the first entry, because that could accidentally get interpreted as a "type" string?

Tricky. If it ends up possible that non-header messages have strings at tag 1, then we have to interpret an unknown type string in the same way we interpret no header at all. We *could* add some magic numbers to the schema, except that wouldn't be backwards compatible. Might not be a huge deal.

> Do we need to set a sane max version length (1024 bytes?) to prevent gnarly/weird parsing?

Is there a way to put a max string length on protobuf fields? @mafintosh

proposals/0000-hypercore-header.md
pfrazee: Ignore, probably. I'll add a note.

proposals/0000-hypercore-header.md
pfrazee: I'm okay with some conventions but I'm not sure I want to standardize a syntax. Just not sure it's worth it. Reason being:

> Let's also say that the version is optional but recommended ("hyperdrive-v0"), and that the x- or X- prefixes are interpreted as "experimental" (aka, no DEP exists yet)

This is totally reasonable, but you then need to interpret `"hyperdrive-v0" == "hyperdrive"`, and I just don't know if we need that level of detail for this.

proposals/0000-hypercore-header.md
pfrazee: Okay I'll rewrite with that in mind.

proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,108 @@
...
+# Reference Documentation
+[reference-documentation]: #reference-documentation
+
+The header message should be written as the first entry of the hypercore. It is not required that all hypercores possess a header.
+
+The header message has the following protobuf schema:
+
+```protobuf
+message HypercoreHeader {
+  required string type = 1;
```
pfrazee: I'm starting to wonder if we should call this `dataStructure` instead of `type`, because "type" is a really generic word that could be used elsewhere in the stack while "data structure" is very specific to what this is.

proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+# Usage Documentation
+[usage-documentation]: #usage-documentation
+
+The "header" is the first entry in a hypercore. It includes a `type` and an optional `extension` which can contain data-structure-specific data.
+
+A program that is reading a hypercore will examine the `type` to determine how to process the hypercore.
+
+```
```
yoshuawuyts: file type here might be helpful to read the rendered version

proposals/0000-hypercore-header.md
```diff
@@ -0,0 +1,108 @@
...
+
+The Dat protocol supports a variety of data structures on top of the hypercore append-only logs, including key-value stores (hyperdb) and file archives (hyperdrive). In order for a program to read the data structure within a hypercore, it must know which structure it is reading. This DEP specifies a "header" entry which can be placed at the first entry of a hypercore in order to provide identifying information. This header also supports custom data for the data-structure, and may be expanded in future DEPs to provide additional standard data.
+
+
+# Motivation
+[motivation]: #motivation
+
+At time of writing, Dat's standard data structures are being expanded upon (hyperdb) while existing data structures are being updated with breaking changes (hyperdrive). In order for these changes to be deployed smoothly to the network, it's important that the following needs be met:
+
+ 1. Programs should correctly identify the data structure type and version within a hypercore. This will enable the program to use the correct handling code (eg hyperdb, hyperdrive-v1, hyperdrive-v2).
```
yoshuawuyts: the example of `hyperdrive-v1` is somewhat at odds with the example of `hyperdrive` throughout the rest of the article. Might be best to pick one and stick with it?

yoshuawuyts:
Are the proposed version names also part of the DEP, or do they only serve as an example? If they only serve as an example, it might be good to make it extra clear that the names aren't being specified in this DEP.

proposals/0000-hypercore-header.md
pfrazee: Agree

proposals/0000-hypercore-header.md
pfrazee: We intentionally are trying to write pseudo code

pfrazee:
> Are the proposed version names for hyperdrive / hyperdb also part of the DEP, or do they only serve as an example? If they only serve as an example, it might be good to make it extra clear that the names aren't being specified in this DEP.

Agree. They're just examples

proposals/0000-hypercore-header.md
mafintosh: Agreed that we should change it to `hyperdrive`

proposals/0000-hypercore-header.md
mafintosh: Noting that this can be updated without any backwards compat issues because of protobuf

bnewbold: `type` is very generic. I'd bikeshed to `contentType` or `contentSchema`, but loosely held opinion (`dataStructure` works for me).

bnewbold:
To summarize the current state of this in my eyes:

- needs a couple tweaks to naming and examples in the DEP text, but I would approve to merge trusting that Paul will make the changes first
- I'd like to move forward, but with the expectation that we should code this up ("working code") and make sure that it allows beaker/dat to disambiguate and handle both old and new hyperdrive before 100% commiting to this scheme (aka, move to Draft ASAP, also prove this out in code to get to Active status ASAP, as getting this correct for the long-term is crucial for the ecosystem)

pfrazee:
Just fixed that "hypercore-v1" thing and added a note that the strings will be in the data structures' DEPs.

I renamed `type` to `dataStructureType`. I'm open to discussion on this. I have some code that's trying to handle dats generally in preparation for more types. In it, I referred to each general item as a "dat" and the instance of hypercore/hyperdrive/hyperdb as that dat's "data structure". So:

```js
var dat = {
  key: '...',
  dataStructureType: 'hyperdrive',
  dataStructure: hyperdrive(...)
}
```

The word `dataStructure` is a little clunky but it does seem to work alright.

PR #41, opened 2018-09-07
Added functions for sign/verify to DEP 0002

### LINKED ISSUE #39: Clarify when and how signatures are generated for hypercore DEP
The hypercore DEP (0002) contains some useful [example code](https://github.com/datprotocol/DEPs/blob/master/proposals/0002-hypercore.md#hash-and-signature-functions) for how nodes are hashed.

I think it's important to do the same thing for signing these nodes because it was a little hard to follow the internals in hypercore and hypercore-crypto to find [this line](https://github.com/mafintosh/hypercore/blob/master/index.js#L1124).

RangerMauve:
Closes #39 

It wasn't instantly obvious that the hash of the roots was being signed when looking at the crypto functions, so I added the `sign` and `verify` psudocode to make it more clear for new implementations.

### DIFF
--- a/proposals/0002-hypercore.md
+++ b/proposals/0002-hypercore.md
@@ -290,9 +290,7 @@ function sign (roots, secretKey) {
   return ed25519.detached.sign(hash, secretKey)
 }
 
-function verify (roots, signature, publicKey) {
-  var hash = root_hash(roots)
-  
+function verify (hash, signature, publicKey) {  
   return ed25519.detached.verify(signature, publicKey)
 }
 ```
--- a/proposals/0002-hypercore.md
+++ b/proposals/0002-hypercore.md
@@ -291,7 +291,7 @@ function sign (roots, secretKey) {
 }
 
 function verify (hash, signature, publicKey) {  
-  return ed25519.detached.verify(signature, publicKey)
+  return ed25519.detached.verify(hash, signature, publicKey)
 }
 ```
 
--- a/proposals/0002-hypercore.md
+++ b/proposals/0002-hypercore.md
@@ -283,6 +283,18 @@ function root_hash (roots) {
 
   return blake2b(buffers)
 }
+
+function sign (roots, secretKey) {
+  var hash = root_hash(roots)
+  
+  return ed25519.detached.sign(hash, secretKey)
+}
+
+function verify (roots, signature, publicKey) {
+  var hash = root_hash(roots)
+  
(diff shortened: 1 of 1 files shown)


pfrazee:
I'm 👍 to this, @mafintosh is this correct?

mafintosh:
LGTM

RangerMauve:
@emilbayes Is that better? :D

emilbayes:
@RangerMauve Sorry, meant on the verify function :) You cannot verify a signature without the original message and the associated public key

RangerMauve:
I think it doesn't matter too much in this case, actually. I think people can infer what's up by seeing how it's the hash from the roots that's being signed.

bnewbold:
(This comment isn't a block or veto of the PR, just a suggestion)

My preference would be to use english or pseudo-code rather than Javascript, despite the existing Javascript examples. My reasoning is we don't want to enshrine particular quirks of Javascript libraries or syntax "as it is today" by accident. For instance, is the "detatch" bit an official recommendation? If not, we should clarify, or better, just not include any specific implementation. The existing code snippets were included to demonstrate routines which were hard to explain in English, which I don't think is the case here. 

The existing "Root hash signatures" section currently says:

> Any time a new root hash is generated, it is signed using the private key. This signature is distributed with the root hash to provide verification of its integrity.

Under "Hash and signature functions":

> The signatures are ed25519 with the sha-512 hash function.

@RangerMauve were those the unclear parts? Could we clarify those paragraphs instead of adding code?

emilbayes:
The comment about "ed25519 with the sha-512 hash function" I think we could replace with something more specific such as the EdDSA RFC, which is also why I suggested dropping the "detached" part: https://tools.ietf.org/html/rfc8032

RangerMauve:
@bnewbold 

I don't think the paragraphs need to be changed, but I still believe that adding the sign and verify functions is useful because with it you can see all the crypto-related pieces in one place which I think will be useful for implementors.

Regarding the `detach` part, I think it's necessary, because as @emilbayes mentioned, libsodium appends the signature to the content by default instead of just returning the signature. The signature is sent separately, so I think it's important to note that. I think that it's safe to assume that implementations are likely to use libsodium as it's what's being used under the hood already.

Regarding using code other than JavaScript, what do you have in mind? The code as it stands doesn't use any javascript-specific features. I think that adding type signatures would be useful, however.

bnewbold:
Thanks for your patience @RangerMauve !

PR #43, opened 2018-10-02
Hypercore Split Resolution DEP

pfrazee:
I had a new idea for solving Hypercore forks that I wanted to propose. From the summary:

 > The Hypercore data-structure is an append-only log which depends on maintaining a linear and unbranching history in order to encode the state-changes of its data content. A "split" event occurs when the log branches, losing its linear history. This spec provides a process for resolving "splits" in a Hypercore log.

It's a fairly simple premise:

> Hypercore's APIs will provide a method for registering a "split handler." The split handler will be implemented by the application using the Hypercore. It may have a standard definition provided by a higher-level data structure such as HyperDB or Hyperdrive.

In practice, it would look something like:

```js
var log = hypercore(..)
log.setSplitHandler((seq, blockA, blockB) => {
  var parsedA = parse(blockA)
  var parsedB = parse(blockB)
  if (parsedA.mtime < parsedB.mtime) return 1
  if (parsedB.mtime < parsedA.mtime) return -1
  return Buffer.compare(blockA, blockB)
})
```

Basically, I think we can trade immutable history and the risk of total dat corruption for (in error conditions) mutable history and the risk of some lost writes.

### DIFF
--- a/proposals/0000-hypercore-split-resolution.md
+++ b/proposals/0000-hypercore-split-resolution.md
@@ -1,0 +1,73 @@
+
+Title: **DEP-0000: Hypercore Split Resolution**
+
+Short Name: `0000-hypercore-split-resolution`
+
+Type: Standard
+
+Status: Draft (as of 2018-10-02)
+
+Github PR: (add HTTPS link here after PR is opened)
+
+Authors: Paul Frazee
+
+
+# Summary
+[summary]: #summary
+
+The Hypercore data-structure is an append-only log which depends on maintaining a linear and unbranching history in order to encode the state-changes of its data content. A "split" event occurs when the log branches, losing its linear history. This spec provides a process for resolving "splits" in a Hypercore log.
+
+
+# Motivation
+[motivation]: #motivation
+
+Hypercore requires a linear history in order to function correctly. A "split" event is a fatal corruption: peers which encounter a split will stop replication, causing the Hypercore to lose utility.
+
+In cases with a strict security requirement this might be useful, but the append-only invariant can be difficult to maintain for users who migrate their dats between devices or restore from backups. For most users, it would be more desirable to risk losing some writes than to risk losing an entire dat due to a split.
+
+This DEP specifies a process for recovering from splits so that users can safely backup or transfer their dats.
+
+
+# Usage Documentation
+[usage-documentation]: #usage-documentation
+
+Hypercore's APIs will provide a method for registering a "split handler." The split handler will be implemented by the application using the Hypercore. It may have a standard definition provided by a higher-level data structure such as HyperDB or Hyperdrive.
+
+The "split handler" function will receive the `seq` number, `blockA`, and `blockB`. It will be expected to return `-1` to accept `blockA` or `1` to accept `blockB`. If `0` is returned, no resolution occurs and the new block is rejected (the current behavior). When a split is resolved, any subsequent messages after the rejected block will also be rejected.
+
+It's expected that the split handler will read the contents of the blocks in order to make decisions about a split. In the case of Hyperdrive, for instance, a split might be resolved using the timestamp of the write.
+
+
+# Reference Documentation
+[reference-documentation]: #reference-documentation
+
+A "split" is defined as a Hypercore which has two or more valid messages with the same sequence number.
+
+"Split resolution" will occur during replication when a "split" is detected due to a received block. Split resolution should not occur during local writes. (Conflicting local writes should be rejected outright).
+
+The split handler must be constructed in such a way that all peers will come to the same decisions independently. Peers should not make a decision based on information such as "time that the block was received," since it is not global knowledge. By only using global information, a split can be reliably resolved across the network.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+"Split resolution" will cause data loss as some part of the history must be discarded. If the managing software is not careful, this can result in massive data loss (e.g. if the split occurs at the first message during recovery). To limit this potential, the managing software can query the network for the latest history in cases where a split is likely (such as a backup recovery process).
+
+"Split resolution" will cause the append-only invariant of Hypercore logs to be optional. This means that file history and versions will not be immutable.
+
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+- Previously we've discussed a "major version pointer" which made it possible possible to recover from a split by publishing a new history ([discussion](https://github.com/datprotocol/DEPs/issues/31)). The "major version pointer" is a less efficient solution as it requires a wholly new history to recover, had its own edge-cases which were difficult to recover from, and was rejected for being too complex.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+
+# Changelog
+[changelog]: #changelog
+
+- 2018-10-02: First complete draft submitted for review
+


pfrazee:
Hypercore is a mechanism for publishing data logs which represent a single user. It is not designed to support global consensus on state, and so neither decentralized consensus or BFT are part of our requirements here.

More broadly, Dat is not designed to support a cryptocurrency. It’s a data sharing and publishing network. It sacrifices global strict/high consistency and atomic global transactions in order to be more efficient.

As a result, we can resolve hypercore splits using only information provided by the hypercore author. We simply need resolution protocols which fit the data structures being encoded on the hypercore, which this proposal enables.

proposals/0000-hypercore-split-resolution.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+- Previously we've discussed a "major version pointer" which made it possible possible to recover from a split by publishing a new history ([discussion](https://github.com/datprotocol/DEPs/issues/31)). The "major version pointer" is a less efficient solution as it requires a wholly new history to recover, had its own edge-cases which were difficult to recover from, and was rejected for being too complex.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
yoshuawuyts: On line 34 the following statement is made:

> It may have a standard definition provided by a higher-level data structure such as HyperDB or Hyperdrive.

Perhaps this would be worth mentioning as part of the unresolved questions?

proposals/0000-hypercore-split-resolution.md
```diff
@@ -0,0 +1,73 @@
...
+
+This DEP specifies a process for recovering from splits so that users can safely backup or transfer their dats.
+
+
+# Usage Documentation
+[usage-documentation]: #usage-documentation
+
+Hypercore's APIs will provide a method for registering a "split handler." The split handler will be implemented by the application using the Hypercore. It may have a standard definition provided by a higher-level data structure such as HyperDB or Hyperdrive.
+
+The "split handler" function will receive the `seq` number, `blockA`, and `blockB`. It will be expected to return `-1` to accept `blockA` or `1` to accept `blockB`. If `0` is returned, no resolution occurs and the new block is rejected (the current behavior). When a split is resolved, any subsequent messages after the rejected block will also be rejected.
```
yoshuawuyts: Using numbers might make a lot of sense for the JS implementation, but less so in Rust (an Enum seems like a better fit). Could the language be adjusted to reflect that better?

proposals/0000-hypercore-split-resolution.md
```diff
@@ -0,0 +1,73 @@
...
+
+The split handler must be constructed in such a way that all peers will come to the same decisions independently. Peers should not make a decision based on information such as "time that the block was received," since it is not global knowledge. By only using global information, a split can be reliably resolved across the network.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+"Split resolution" will cause data loss as some part of the history must be discarded. If the managing software is not careful, this can result in massive data loss (e.g. if the split occurs at the first message during recovery). To limit this potential, the managing software can query the network for the latest history in cases where a split is likely (such as a backup recovery process).
+
+"Split resolution" will cause the append-only invariant of Hypercore logs to be optional. This means that file history and versions will not be immutable.
```
yoshuawuyts: What are the implications of changing the file history and versions to not be immutable? Which assumptions will be changed by this? It feels like this would be a pretty big change to the core premise of Hypercore.

proposals/0000-hypercore-split-resolution.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Rationale and alternatives
+[alternatives]: #alternatives
+
+- Previously we've discussed a "major version pointer" which made it possible possible to recover from a split by publishing a new history ([discussion](https://github.com/datprotocol/DEPs/issues/31)). The "major version pointer" is a less efficient solution as it requires a wholly new history to recover, had its own edge-cases which were difficult to recover from, and was rejected for being too complex.
```
yoshuawuyts:
```suggestion
- Previously we've discussed a "major version pointer" which made it possible to recover from a split by publishing a new history ([discussion](https://github.com/datprotocol/DEPs/issues/31)). The "major version pointer" is a less efficient solution as it requires a wholly new history to recover, had its own edge-cases which were difficult to recover from, and was rejected for being too complex.
```

proposals/0000-hypercore-split-resolution.md
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+Hypercore requires a linear history in order to function correctly. A "split" event is a fatal corruption: peers which encounter a split will stop replication, causing the Hypercore to lose utility.
+
+In cases with a strict security requirement this might be useful, but the append-only invariant can be difficult to maintain for users who migrate their dats between devices or restore from backups. For most users, it would be more desirable to risk losing some writes than to risk losing an entire dat due to a split.
+
+This DEP specifies a process for recovering from splits so that users can safely backup or transfer their dats.
```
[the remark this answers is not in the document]
pfrazee: It's a trade of costs. Currently, a split-event results in total hypercore corruption: no more forward progress can be made in the hypercore, and it must be replaced with a new keypair and history. This proposal provides a solution for resolving splits, therefore it re-enables forward progress on the hypercore. However it does not create a process for restoring any data in the "discarded branch" and so it leaves the potential for data loss.

The reason that this proposal doesn't include some form of data restoration from the dead-branch is that hypercore's semantics are too generic to create a universal solution. A restoration process is possible, but it would depend on the data-structure encoded on the hypercore. For instance, a hyperdrive might handle restoration by showing the user a list of the orphaned files, and prompting them to either "restore" or "discard" each file. This would cause each file to be rewritten onto the newly live branch.

proposals/0000-hypercore-split-resolution.md
pfrazee: It's not really an unresolved question; it's more a prompt for followup DEPs providing the standard definitions for the various data structures.

proposals/0000-hypercore-split-resolution.md
pfrazee: Sure

proposals/0000-hypercore-split-resolution.md
pfrazee: The largest implication is that specifying a revision-number no longer provides a strong guarantee of content. Explaining in more detail:

With immutable history, it's enough to provide the pair `(pubkey, revision)` to specify a set of content with a strong guarantee of the content. That is, for any `(pubkey, revision)` pair, there is only one dataset which may exist in the world. (That's assuming that the history's immutability is actually maintained, which depends on the network gossiping effectively about any known splits.)

With mutable history, to provide a strong guarantee of content, you must include a content hash. Therefore you must provide a triple `(pubkey, revision, hash)` to specify a strong guarantee.

We had already planned to create the triple-form as something called "strong links" for two reasons:

 1. When DNS is involved, you're actually specifying the pair `(domain-name, revision)` which means that the immutable history guarantee is not upheld. DNS is common enough that this is a concern.
 2. It's not physically impossible to maintain a split in the network, at least for some time, making the immutable history guarantee somewhat weak.

pfrazee: Another implication has to do with processing guarantees. For instance, if we assert immutable history, then a process which ingests hypercores and produces computed views can assume that previously-processed revisions will never change. This is a nice optimization which might even enable the process to discard revisions after processing them.

If the immutable history guarantee is lost, that optimization is no longer possible, and the processor has to watch for splits and recompute its views accordingly.

proposals/0000-hypercore-split-resolution.md
yoshuawuyts: Gotcha.

pfrazee:
@mafintosh and I discussed this in a WG meeting. We came to two conclusions:

 1. This spec is missing an important facet -- it needs to ensure that the hypercore owner is made aware of the full split prior to attempting resolution. When a split occurs, it's because the owner isn't aware of some previously-published history. It will need to query peers in order to discover the starting-point of the split. And, if it wants to restore any data from the split, it's going to need to download as much history as possible from the discarded branch. Currently this spec doesn't cover how that can happen.
 2. We have a pretty full docket with the hyperswarm and hyperdrive v2 updates, so this spec probably needs to wait till those are done.

PR #49, opened 2018-11-17
Amending reservations against this DEP

martinheidegger:
To me this DEP is quite problematic. It has the potential to be of major impact on how we understand DAT and can degrade the advantages given by DAT a lot, depending on how it is used in practice. This PR adds my reservations against this proposal.

### DIFF
--- a/proposals/0006-session-data-extension.md
+++ b/proposals/0006-session-data-extension.md
@@ -42,14 +42,15 @@ The client may respond to the message by emitting an event, so that it may be ha
 
 After publishing this DEP, the "Beaker Browser" will implement a Web API for exposing the `'session-data'` protocol to applications. It will restrict access so that the application code of a `dat://` site will only be able to set the session data for connections related to its own content.
 
-
 # Drawbacks
 [drawbacks]: #drawbacks
 
 - This DEP may present privacy concerns, as it may be used to track users in a similar fashion to HTTP Cookies.
 - The payload of the `'session-data'` message is not authenticated in any way. If a public key is sent, proof of ownership of the private key is not provided. The lack of trust must be considered by applications which leverage the data.
 - If the recipient of the `'session-data'` message is not authenticated (as is currently the case in all Dat replication connections) the client will not know who is receiving the payload and may broadcast sensitive information.
-
+- By implementing this DEP, servers **can** return different data depending on the session-data as such we must expect that they will. This is hidden behavior that can not be explained without debugging the server within the session state: Users need to be educated about this fact.
+- Different data per session can be implemented at the discretion of the server. It may be related to the session data or to unrelated data (time, amount of other users, etc.). The client may never share any data received within a session. To be consistent: every new session needs to be treated as independent from former data received.
+- In order to allow users to enter session data, clients need to implement an UI for this as well: if the session data is only machine readable: UI's may be difficult to implement.
 
 # Rationale and alternatives
 [alternatives]: #alternatives
@@ -62,6 +63,7 @@ An alternative approach would be to establish an ephemeral messaging channel, pe
 # Changelog
 [changelog]: #changelog
 
+- 2018-11-15: Drawback information amended 
 - 2018-06-06: Merged as draft
 - 2018-05-31: First complete draft submitted for review
 


pfrazee:
Could servers return different data depending on the session-data? That would create a detectable hypercore split, wouldn’t it?

martinheidegger:
I can think of only three ways a server can react to input, if neither of this is done: the server doesn't need session data imo.

- delay/increase data throughput (prioritizing)
- restricting access to immutable data (HAVE respond with different sections.)
- change the response.

Changing the HAVE sections depending on input response seems like the obvious thing to attempt and it can be done by simply returning different hypercore data to different sockets. If the client doesn't try to reshare the DAT (which is the default in beaker),  no problem will occur. I believe it should be possible for two entirely different trees to be returned.

bnewbold:
I agree that this DEP changes the semantics of what the Dat/hypercore protocol is used for, and that is can have a "major impact on how we understand DAT", but the history here is that these changes were already happening (with Cabal), and this DEP was an attempt to adapt to those changes already happening.

> servers **can** return different data depending on the session-data as such we must expect that they will

I don't see this as a strong argument against this DEP, as "servers" can **already** do or return whatever they want, based on things like "client" IP address, latency, peer ID, etc.

I think it would be pretty clearly against the dat/hypercore semantics to return different hypercore feed content (to be specific, feed entries with the same index number but different hashes, thus splitting/forking the feed) to different users (or user agents). While technically correct that "If the client doesn't try to reshare the DAT (which is the default in beaker), no problem will occur", this would be extremely fragile and I think against other presumptions in the ecosystem. We might be misunderstanding what you mean by "data per session" though.

I think this DEP is enabling three specific things:

- whether to return any feed content at all (aka, access control), either for the first feed connected, or any ancillary feeds (additional channels). This is related to your second bullet point, but at the FEED/channel level, not the HAVE/entry level
- whether the "server" should accept or request (via FEED message) additional feeds, for which the connecting "client" has the signing key. AKA, to allow the "client" to "push" content (entire feeds) to the "server"
- allowing the "server" to keep track of connection state, which is sort of a side-band to some use cases of Dat (synchronizing data and static content), but important for real-time/interactive/collaborative use cases

Note: I continued using the client/server language this thread was started will, but we usually call all computers "peers" in the network.

martinheidegger:
> I don't see this as a strong argument against this DEP, as "servers" can already do or return whatever they want, based on things like "client" IP address, latency, peer ID, etc.

In practice I do not believe that IP address, latency or peer-id is reliably to identify a client. Particularly if the use of proxies or backups becomes widespread. On top of that user-data basically practically begs for different response per user.

> This would be extremely fragile and I think against other presumptions in the ecosystem. 

Yes, totally agree: an implementation would break the DAT ecosystem, that is one of the reason I added my reservations: "because we can" ruins many systems: I believe that is the case here: server implementer will use this approach because they can.

> whether to return any feed content at all (aka, access control), either for the first feed connected, or any ancillary feeds (additional channels). 

I am second-guessing here if this will work in practice: once a user downloaded a feed, the client-software might very-well immediately share it on the network _(and giving access to all other peers)_.

The only case where this will work is if the client does-not re-share the DAT. Particularly in that case: different feeds - or personalized data - becomes more likely/viable.

> Note: I continued using the client/server language this thread was started will, but we usually call all computers "peers" in the network.

I used the "server"/"client" distinction here because I can not imagine a user facing software that can properly manage session-data. Also I am wondering if this DEP doesn't mean that there will be actually DAT servers & clients in future.

pfrazee:
> I used the "server"/"client" distinction here because I can not imagine a user facing software that can properly manage session-data. Also I am wondering if this DEP doesn't mean that there will be actually DAT servers & clients in future.

The example use-case for this DEP was for peers to identify themselves in a chat room.

martinheidegger:
> The example use-case for this DEP was for peers to identify themselves in a chat room.

I assume this is for the use-case that a user should be shown as "online/offline": How would this be implemented for the whole chat room? The user is connected to a subset of peers for this chat-room: By sending an identification to his direct peers, the further-away-peers will not be notified: showing this user offline even though he is actually connected to the chat. It seems to me like "online/offline" becomes a state that should be shared between the peers of the chat client and as such would be better suited as example for content rather than session data?!

PR #53, opened 2018-12-12
Intent DEP

martinheidegger:
_(Note: this is the first DEP I authored, mentoring welcome!)_

I had the thought expressed in this DEP a while back but when working on DAT Desktop today I remembered it, so I wrote it down.

This DEP - when looked at in some light - allows for some heterogeneity in the
DAT network which feels good to me.

### DIFF
--- a/proposals/0000-intents.md
+++ b/proposals/0000-intents.md
@@ -1,0 +1,87 @@
+Title: **DEP-0000: Intents**
+
+Short Name: `0000-intents`
+
+Type: Standard
+
+Status: Draft (as of 2018-12-12)
+
+Github PR: [Draft](https://github.com/datprotocol/DEPs/pull/53)
+
+Authors: [Martin Heidegger](https://github.com/martinheidegger)
+
+
+# Summary
+[summary]: #summary
+
+DAT clients can have various purposes when connecting to other peers. A _"backup"_-peer might have the intention to store and
+seed as many much of a given DAT as possible, while a _"download-latest-once"_-peer might only try to download the
+latest version and leave the network when its done and still a _"browse"_-peer might be only interested in parts of a DAT
+at a time a human points at something.
+
+This DEP specifies an additional, optional [Handshake property][Handshake-properties] _(specified in [#8][wireprotocol-pr])_
+to announce the intent of the peer.
+
+[Handshake-properties]: https://github.com/pfrazee/DEPs/blob/e761143a051dc9735364d8897baf691bd2ed9d68/proposals/0000-wire-protocol.md#handshake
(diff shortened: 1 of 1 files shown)


proposals/0000-intents.md
```diff
@@ -0,0 +1,87 @@
...
+- `browse`: A browsing client that intends to download and dismiss parts of the downloaded structure at free will.
+- `browse-seed`: Similar to `browse` but with the intention to re-seed the downloaded parts.
+- `backup`: A client that downloads and re-seeds as many versions of an given hypercore as specified.
+- `seed`: A client that downloads and re-seeds only one version _(usually the latest)_ of an hypercore.
+- `author`: A client that has the secret keys to author that hypercore.
+- `publish`: Similar to `seed` but with the intent also publish outside of the DAT protocol _(i.e. a HTTP service)_
+- `backup-publish`: Similar to `backup` but it intends to publish outside of the DAT protocol _(i.e. a HTTP service)_
+- `proxy`: A client that connects in order for re-share whatever is requested to further unspecified clients.
+
+For sustom strings, the recommendation is to use `non-standard-` as a prefix for intents in order to make sure
```
dkastl: Should be `custom`.

pfrazee:
Yeah this is an interesting idea. @mafintosh has been exploring similar kinds of information in the Hyperswarm DHT to help with connection priorities. For this spec I'd be curious to explore the specific effects of different intents. Eg, for each intent, how are recipients expected to react?

martinheidegger:
There should be no necessity of change in behavior by the recipients - connection-wise. I think adding it at this point in time would be premature.

What I think might be interesting to specify is how to highlight this in a User-interface, while I have ideas on it _(like showing an eye-icon when there is a browser but no seed / download)_. Would that be enough for you?

martinheidegger:
Hinted at by @karissa _(thx!)_: Intents could be very important to visualize the state of backups of a DAT: Right now hypercores do not ask other peers to share all their "have"'s because that would be way-too-much data to be sent through the network. However, having dedicated "backup" peers would make the selection easier: if a "backup" peer connects, the desktop could assume that it always "wants" to know everything that the desktop has. It could also be asked to share all "have"'s in order to identify which versions have been replicated on a "backup" peer.

karissa:
I think this is useful but also introduces some metadata which might not be desired for certain use cases, especially where high security is needed. So I think it is important it stays optional

martinheidegger:
Furthermore I came to think that it can be used for protocol optimization: If a peer identifies as `browse` (non-seed) then no `WANT` requests should be sent to this peer, as its have can be assumed that this peer doesn't intend to upload any `HAVE`.

PR #64, opened 2019-11-19
Fix typo

decentral1se:


RangerMauve:
@pfrazee @mafintosh Any chance this could get merged / can we get more collaborators in this repo so we can merge PRs like this without bugging you? 💜

pfrazee:
Yeah I can merge - I think the idea is that the WG members are the only ones allowed to have merge rights, so we'll have to dig into that later

RangerMauve:
Might be good to dig into it sooner given that the WG hasn't had a meeting in months and most of the members are inactive. :P

PR #67, opened 2020-11-23
fix padding size math

zxcat:
doc says padding is 21 bytes for signatures and tree, but it should be 17:
```
      <3 byte magic string: 0x050257>
      <1 byte header type: 0x02>
      <1 byte version number: 0>
      <2 byte entry size: 40>
      <1 byte algorithm name length prefix: 7>
      <7 byte algorithm name: BLAKE2b>
      <17 byte padding>
```
3+1+1+2+1 + 7+**17** = 32


martinheidegger (approved):
Seems about right 😉

RangerMauve:
ping @mafintosh @pfrazee

mafintosh:
👍

